PR #28, opened 2021-11-21
Refactor GPTQ, LUT and adding initial support for mixed precision 

reuvenperetz:
General changes:
 Updated version number to 1.1.0
 Refactored facade: Used new refactored files of quantizing the graph and bias correction, extracted code to internal functions for the facade use, removed logging to Tensorboard with bops computation.
 Added printing in logger methods.
 Added similarity_analyzer with implementations for different similarity metrics (MSE, CS, etc.). The functions are aligned to return low values for similar tensors and high values for a high disparity.
 Renamed functions in kl_selection, lp_selection, mae_selection, mse_selection and migrate error computation to new similarity_analyzer.
 Renamed symmetric_uniform quantization with power_of_two quantization.
 Changed shift_negative_ratio in quantization configuration from 0.25 to 0.05.
 Removed node_to_bops and its use in the TensorboardWriter class, so resources on TensorBoard will not contain the bops computation.
 Renamed threshold_selection package to quantization_params_generation and moved it under quantization package.
 Removing unused imports.
 Split bias correction logic into 2 files: computation of the bias and applying the correction term.
 Added new common constants for node's shared attributes names and a multiplier number of bits for kmeans-LUT quantizer.
 Added mapping from Keras layers to their kernel attributes names to quantize in FrameworkInfo.
 Added method for the base graph to return a topologically sorted list of its nodes.
Removed the field weights_keys from Node (instead, use the weights dictionary the node already holds).
 Updated compare_points in graph_info to return nodes topologically sorted.
 Added PuLP to requirements file for linear programming in mixed-precision.
 Copied graph before applying substitutions in substituter.
 Added Dense to MarkActivation substitution.
 Removed cosine_similarity and tensor_norm from nn_visualizer to an outer package.
 Added function for getting comparing points in nn_visualizer.
 Added new feature networks tests.

Issues:
 Fixed bug in TensorboardWriter when a node has a shape of ().
 Fixed node builder: Mark the "base" node in case of a reused group with its reused group, so we can skip it during substitutions.
 Fixed model reader to build a list of tensors references when the tensors are in a tuple (and not only a list).
 Fixed bug in BatchNormalization folding substitution when the linear node is reused.
 Fixed bug in decomposing SeperableConv2D folding substitution when the linear node is reused.
 Fixed bug in channels equalization when bias is None at first layer we substitute.
 Fixed missing license.

Documentations:
 Updated README by moving the site link to the first paragraph.

Mixed-Precision:
 Split node's weights quantization configuration into 2 kinds of quantization configuration: "candidates" and final. The candidates are the possible configurations during the mixed-precision search, and each node's "final" configuration is chosen as one of the candidates.
 Split activation_weights_quantization in Node into weights and activations different methods.
 Split ChangeWeightsQuantConfigAttr into: ChangeCandidatesWeightsQuantConfigAttr and ChangeFinalWeightsQuantConfigAttr.
 Splited ChangeQuantizationMethod into: ChangeFinalWeightsQuantizationMethod and ChangeActivationQuantizationMethod.
 Added Node class the method to return a "unified" version of its properties (to consider all candidates configurations).
 Added bit_width_setter file to set a mixed-precision configuration to a graph (set final configurations).
 Added distance_weighting file with functions to return weights for weighting distances in the mixed-precision package.
 Added KPI for different constraints when searching a mixed-precision configuration.
 Added mixed-precision quantization configuration the user can pass to the mixed-precision facade.
 Added facade to the search package in mixed-precision.
 Added mixed-precision search manager class to include all information needed and operate the different parts of the search process.
 Added an integer linear programming search for mixed-precision configurations.
 Added mixed-precision final configuration to the return UserInformation object.
 Updated model builder modes with MIXEDPRECISION mode to build a dynamic quantize Keras model.
 Added PuLP to requirements file for linear programming in mixed-precision.
 Added sensitivity evaluation in mixed_precision package (in keras package) to compute a distance of a mixed-precision configuration from a baseline model (the float model).
 Added QuantizeConfig factory to mixed_precision package in keras.
 Added SelectiveQuantizer class holds kernel and its' quantized versions in different bit widths and returns the desired one on the fly.
 Added SelectiveWeightsQuantizeConfig to wrap a layer in Keras model with a SelectiveQuantizer so the layer can use the kernel in different bit widths during inference.
 Added feature networks tests and base feature test methods for returning KPI and mixed-precision configuration to set.

GPTQ:
 Name change: KD->GPTQ
 Refactored KD QuantizeConfigs generation: Add QuantizeConfig factory to gptq package.
 Added utils.py to gptq package as part of the refactor of the KD package.
 Added new get_trainable_params: only get quantizer params and bias.
 Added fakequant layer after input layer in fxp model to match PTQ model in GPTQ mode.
 Added support of getting loss as input from API (in gptq config, and change default loss to MSE)
 Added quantizer for both weight and activation of a layer in GPTQ.
 Updated update_layer_quantization_params in activation_quantizer_kd_config to return another dictionary.
 Added max_lsbs_change to weight_quantizer_kd_config to allow a maximal change in the number of LSBs of quantized values.

KMEANS:
 Refactored Kmeans and LUT-Kmeans quantizers:
 Added 'eps' when dividing with scales for the zero case.
Added an exception in LUT-Kmeans for a case where clusters are greater than the multiplier number of bits.
 Extracted scale search and cluster centers to new files in quantization_qparams_generation.


model_compression_toolkit/common/quantization/quantizers/power_of_two_quantizer.py
```diff
@@ -42,7 +42,7 @@ def symmetric_uniform_quantizer(tensor_data: np.ndarray,
     """
     threshold = quantization_params.get(THRESHOLD)
     if threshold is None:
-        raise Exception('\'weights_threshold\' parameter must be defined in \'quantization_params\'')
+        raise Exception("'weights_threshold' parameter must be defined in 'quantization_params'")
```
haihabi, line 45: Why this is true weights thresholds?
It should be threshold no?

model_compression_toolkit/keras/quantization_facade.py
```diff
@@ -208,56 +228,299 @@ def keras_post_training_quantization(in_model: Model,
...
+     For now, mixed precision is supported for weights only.
+     If a gptq configuration is passed, the quantized weights are optimized using gradient based post
+     training quantization by comparing points between the float and quantized models, and minimizing the observed loss.
+
+     Args:
+         in_model (Model): Keras model to quantize.
+         representative_data_gen (Callable): Dataset used for calibration.
+         n_iter (int): Number of calibration iterations to run.
+         quant_config (MixedPrecisionQuantizationConfig): QuantizationConfig containing parameters of how the model should be quantized.
+         fw_info (FrameworkInfo): Information needed for quantization about the specific framework (e.g., kernel channels indices, groups of layers by how they should be quantized, etc.). `Default Keras info <http://bitbucket-rh7:7990/projects/DNNQUAN/repos/posttrainingquantization/browse/network_optimization_package/keras/default_framework_info.py#112>`_
```
haihabi, line 289: Remove bitbucket url

model_compression_toolkit/keras/quantizer/mixed_precision/quantization_config_factory.py
```diff
@@ -0,0 +1,55 @@
...
+from tensorflow.keras.layers import DepthwiseConv2D, Conv2D, Conv2DTranspose, Dense
+from tensorflow_model_optimization.python.core.quantization.keras.default_8bit.default_8bit_quantize_registry import \
+    QuantizeConfig
+
+from model_compression_toolkit import common
+from model_compression_toolkit.keras.constants import DEPTHWISE_KERNEL, KERNEL
+from model_compression_toolkit.keras.quantizer.mixed_precision.selective_weights_quantize_config import \
+    SelectiveWeightsQuantizeConfig
+
+KERNEL_NAME = {Conv2DTranspose: KERNEL,
```
haihabi, line 26: We don"t have this information in the framework info class?

tests/feature_networks_tests/feature_networks/input_scaling_test.py
```diff
@@ -31,8 +31,8 @@ def __init__(self, unit_test):
 
     def get_quantization_config(self):
         return mct.QuantizationConfig(mct.ThresholdSelectionMethod.NOCLIPPING, mct.ThresholdSelectionMethod.NOCLIPPING,
-                                       mct.QuantizationMethod.SYMMETRIC_UNIFORM, mct.QuantizationMethod.SYMMETRIC_UNIFORM,
-                                       16, 16, False, False, True, input_scaling=True)
+                                      mct.QuantizationMethod.POWER_OF_TWO, mct.QuantizationMethod.POWER_OF_TWO,
+                                      16, 16, False, False, True, input_scaling=True)
 
     def create_inputs_shape(self):
         return [[self.val_batch_size, 224, 244, 3]]
```
haihabi, line 38: Reduce size to 16x 16 or 8x8

model_compression_toolkit/common/quantization/quantizers/power_of_two_quantizer.py
reuvenperetz: This function is used for weights quantization only

PR #35, opened 2021-12-04
Refactor to support more frameworks

reuvenperetz:
Removed tensorflow packages from requirements file.
Updated github workflows to use tensorflow packages when needed.
Split tests to two packages: common_tests and keras_tests.
Added new workflow to test the common tests.
Moved GPTQConfig and ModelBuilderMode to common.
Removed GPTQConfig loss default function (as it was relevant to TF only).
Changed MCT init to check for TF package, and only if exists - import TF's facade functions.
Moved substitue function to common (in a new package called substitutions).
Moved Matchers package in common to the new substitutions package in common.
Removed TF importing from: apply_bias_correction_to_graph and from test_histogram_collector.
Added a new abstract class for framework specific implementation (FrameworkImplementation).
Added a class KerasImplementation class which implements the methods of FrameworkImplementation,
to support Keras in MCT.
Created a new post_training_quantization in common to run the entire process of
optimization. This logic moved out from quantization_facade, and quantization_facade
is now only two slim functions that use post_training_quantization.
Moved OutTensor namedtuple from keras to common (from connectivity handler to base graph).

Fix bug in histogram collector: When we computed new bins values when merging the
collected histograms, the maximal tensor value was out of this range. We now
compute new bins values including the maximal tensor value.


model_compression_toolkit/common/base_substitutions.py
```diff
@@ -18,8 +18,7 @@
 from typing import Any
 
 from model_compression_toolkit.common.graph.base_graph import Graph
-from model_compression_toolkit.common.matchers.base_matcher import BaseMatcher
-
+from model_compression_toolkit.common.substitutions.matchers.base_matcher import BaseMatcher
```
haihabi, line 21: Why change the matcher position?

model_compression_toolkit/__init__.py
```diff
@@ -27,4 +27,16 @@
 from model_compression_toolkit.common.defaultdict import DefaultDict
 from model_compression_toolkit.common import network_editors as network_editor
 
+import importlib
+
+tf_spec = importlib.util.find_spec("tensorflow")
+was_tf_found = tf_spec is not None
+
+if was_tf_found:
+    from model_compression_toolkit.keras_framework.quantization_facade import keras_post_training_quantization, \
```
haihabi, line 36: 1. Why change keras to keras_framework ?
2. This verification should be inside the keras_post_training_quantizaiton and if it is called without keras then there is an error message

model_compression_toolkit/common/graph/base_graph.py
```diff
@@ -29,9 +29,11 @@
 from model_compression_toolkit.common.statistics_collector import BaseStatsContainer
 from model_compression_toolkit.common.statistics_collector import scale_statistics, shift_statistics
 from model_compression_toolkit.common.user_info import UserInformation
-from model_compression_toolkit.keras.reader.connectivity_handler import OutTensor
 from model_compression_toolkit.common.logger import Logger
 
+OutTensor = namedtuple('OutTensor', 'node node_out_index')
```
haihabi, line 34: Why this is required?

model_compression_toolkit/common/mixed_precision/bit_width_setter.py
```diff
@@ -25,7 +26,7 @@
 
 
 
-def set_bit_widths(quant_config: MixedPrecisionQuantizationConfig,
+def set_bit_widths(quant_config: QuantizationConfig,
```
haihabi, line 29: Why this should be changed?

model_compression_toolkit/common/post_training_quantization.py
```diff
@@ -17,23 +17,26 @@
 import copy
 import os
 from functools import partial
-from typing import Callable, List, Tuple
+from typing import Callable, List, Tuple, Any
 
 from tensorflow.keras.models import Model
```
haihabi, line 22: Why there still a keras import in this file?

haihabi, line 22: Why there is still a keras import in this file?

requirements.txt
```diff
@@ -1,11 +1,10 @@
-tensorflow==2.5.*
 networkx==2.5
 tqdm
 Pillow
 numpy
 opencv-python
 scikit-image
 scikit-learn
-tensorflow_model_optimization
```
haihabi, line 9: Add a check regarding tensorflow_model_optimization

PR #36, opened 2021-12-05
Refactor support multiple frameworks

reuvenperetz:


.github/workflows/nigthly.yml
```diff
@@ -34,6 +34,8 @@ jobs:
         run: |
           rm -rf model_compression_toolkit
           pip install mct-nightly
+          pip install tensorflow==2.5.*
```
haihabi, line 37: Why not install tensorflow 2.6 ?

PR #45, opened 2021-12-11
Support atomic ops

reuvenperetz:
- Changed Node to BaseNode.
- Removed op_call_args from BaseNode.
- Created a new class that inherits from BaseNode,
called FunctionalNode to represent layers that wrap
functions to call with arguments.
- Changed node_builder to build FunctionalNodes with
their arguments to pass when converting a node back to
its original framework. The arguments do not include the
tensors as we use the tensors we build in model_builder.
- Updated FrameworkInfo methods to consider functional
nodes in activation ops and no_quantization ops.
- Updated model_builder to build functional nodes when
it's the case.
- Updated Keras model reader to use the attribute
'keras_inputs' instead of 'input_tensors' to get node
inputs (except for Input layers, where we still need
to use input_tensors).
- Updated Keras default information with functional ops
in activation ops and no_quantization ops.
- Added feature network tests.

Issues fixed:
- Fixed issue in quantization_analyzer when a node has a
list of NoStatsConllectors at its output, but the next node
is a kernel op and we use a bias correction. The issue was
that we did not update the NoStatCollector list, and during
bias correction, we tried to use its mean (and failed, of course).
- Fixed issue in ModelCollector, where we should not collect
statistics (e.g., tf.split layer) but we still tracked it in
the ModelCollector, which caused a mismatch between the data
we collected and the collectors we have tried to update. Now,
the behavior is to not track a list of collectors that do not
collect data.


model_compression_toolkit/common/graph/functional_node.py
```diff
@@ -0,0 +1,59 @@
+from typing import Dict, Any, Tuple, List
+
+from model_compression_toolkit.common.graph.base_node import BaseNode
+import numpy as np
```
haihabi, line 4: Why there is a need for numpy import?

model_compression_toolkit/keras/back2framework/model_builder.py
```diff
@@ -105,11 +106,14 @@ def build_input_tensors_list(node: Node,
...
+
+
+def run_operation(n: BaseNode,
                   input_tensors: List[List[TFReference]],
                   op_func: Layer,
-                  input_nodes_to_input_tensors: Dict[Node, Any],
-                  mode: ModelBuilderMode = ModelBuilderMode.QUANTIZED) -> List[TFReference]:
+                  input_nodes_to_input_tensors: Dict[BaseNode, Any],
+                  mode: ModelBuilderMode = ModelBuilderMode.QUANTIZED,
+                  fw_info: FrameworkInfo = None) -> List[TFReference]:
```
haihabi, line 116: Remove fw_info

PR #51, opened 2021-12-27
Feature layer tests

reuvenperetz:


tests/test_suite.py
```diff
@@ -33,4 +33,4 @@
 
     # ----------------   Join them together ane run them
     comboSuite = unittest.TestSuite(suiteList)
```
haihabi, line 35: Layer test is missing from test_suite

tests/keras_tests/feature_networks_tests/feature_networks/scale_equalization_test.py
```diff
@@ -45,11 +46,11 @@ def get_quantization_config(self):
                                       weights_per_channel_threshold=True,
                                       activation_channel_equalization=True)
 
-    def create_inputs_shape(self):
+    def get_input_shapes(self):
```
haihabi, line 49: This test fails due to the small batch size.
maybe we should change the compare function? (remove cosine similarity) and add another test?

PR #96, opened 2022-02-15
Pytorch implementation for Mixed Precision

ofirgo:
Mixed Precision implementation for Pytorch framework:

- New wrapper class `PytorchMixedPrecisionWrapper` which allows configuring a Pytroch module (a layer in the model) with different values of bit-width.
- Modify `model_builder` which allows creating a MP model from a graph, such that each configurable layer in the model is wrapped with a `PytorchMixedPrecisionWrapper`.
- Dedicated sensitivity evaluation method for MP with Pytorch framework.
- In addition, functionalities that are shared for both Pytorch and TensorFlow were extracted to a common environment, under the new class `SensitivityEvaluationManager`, which maintains all necessary data and objects for running sensitivity evaluation during MP search.

### DIFF
--- a/model_compression_toolkit/common/mixed_precision/sensitivity_evaluation_manager.py
+++ b/model_compression_toolkit/common/mixed_precision/sensitivity_evaluation_manager.py
@@ -1,0 +1,239 @@
+# Copyright 2022 Sony Semiconductors Israel, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import numpy as np
+from typing import Callable, Any, List
+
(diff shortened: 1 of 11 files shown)


model_compression_toolkit/common/mixed_precision/sensitivity_evaluation_manager.py
```diff
@@ -0,0 +1,239 @@
...
+        with different bitwidth for quantization.
+    Args:
+        graph: Graph to search for its MP configuration.
+    Returns: A sorted list of nodes.
+    """
+    sorted_nodes = graph.get_configurable_sorted_nodes()
+    sorted_configurable_nodes = []
+    for n in sorted_nodes:
+        sorted_configurable_nodes.append(n)
+
```
reuvenperetz, line 153: We have this method for the graph. Why do we need it here as well?

model_compression_toolkit/pytorch/back2framework/model_builder.py
```diff
@@ -121,34 +115,42 @@ class PytorchModelBuilder(torch.nn.Module):
...
         self.append2output = append2output
-        for n in self.node_sort:
-            if not isinstance(n, FunctionalNode):
-                self.add_module(n.name, node_builder(n))
+
+        if mode == ModelBuilderMode.MIXEDPRECISION:
+            for n in self.node_sort:
+                if not isinstance(n, FunctionalNode):
+                    if n.is_weights_quantization_enabled():
+                        self.add_module(n.name, PytorchMixedPrecisionWrapper(n, fw_info))
```
reuvenperetz, line 139: Why don't we take the configurable nodes from the manager?

PR #98, opened 2022-02-20
Added Pytorch features:

lior-dikstein:
- support for models with parameters and constant tensor values
- support for models with torch.Tensor functions
- added relevant tests


tests/pytorch_tests/model_tests/feature_models/parameter_net_test.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    This tests check a model with a parameter which is a constant at inference time.
+    In addition, the model has an addition layer regular constant tensor
+    """
+    def __init__(self, unit_test):
+        super().__init__(unit_test)
+
+    def create_inputs_shape(self):
+        return [[self.val_batch_size, 3, 32, 32]]
+
```
reuvenperetz, line 46: Why don't we use the default shape?

PR #128, opened 2022-04-03
refactor default hw model

ofirgo:
Refactor default hardware model to allow generating new models for tests and experiments, while keeping the framework hardware modeling mapping

### DIFF
--- a/model_compression_toolkit/hardware_models/keras_hardware_model/keras_default.py
+++ b/model_compression_toolkit/hardware_models/keras_hardware_model/keras_default.py
@@ -15,6 +15,8 @@
 
 import tensorflow as tf
 
+from model_compression_toolkit.common.hardware_representation import HardwareModel
+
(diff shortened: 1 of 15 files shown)


model_compression_toolkit/hardware_models/keras_hardware_model/keras_default.py
```diff
@@ -57,7 +71,7 @@ def get_default_hwm_keras():
 
         hwm.OperationsSetToLayers("AnyReLU", [tf.nn.relu,
                                               tf.nn.relu6,
-                                              hwm.LayerFilterParams(ReLU, negative_slope=0.0),
+                                              ReLU,
```
reuvenperetz, line 74: I think I mislead you with that. It should stay unchanged

reuvenperetz (changes requested):

PR #132, opened 2022-04-04
Get n_bits candidates from hw model

ofirgo:
Refactor to allow getting quantization n_bits configuration candidates from hardware model instead of quantization config.
Main changes include:

- Remove n_bitsfor both weights and activation from QuantizationConfig and remove n_bits_candidates from MixedPrecisionQuantizationConfig (refactor entire lib accordingly)
- Set node's candidates config with info from hardware model
- Large refactor in tests - to adjust setting test configuration with hw model

### DIFF
--- a/tests/keras_tests/feature_networks_tests/feature_networks/nested_networks/nested_model_multiple_inputs_test.py
+++ b/tests/keras_tests/feature_networks_tests/feature_networks/nested_networks/nested_model_multiple_inputs_test.py
@@ -16,6 +16,9 @@
 import model_compression_toolkit as mct
 import tensorflow as tf
 
+from model_compression_toolkit.hardware_models.keras_hardware_model.keras_default import generate_fhw_model_keras
+from tests.common_tests.helpers.generate_test_hw_model import generate_test_hw_model
+
 if tf.__version__ < "2.6":
     from tensorflow.python.keras.engine.functional import Functional
     from tensorflow.python.keras.engine.sequential import Sequential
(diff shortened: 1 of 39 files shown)


tests/pytorch_tests/model_tests/base_pytorch_test.py
```diff
@@ -34,21 +36,34 @@ def __init__(self, unit_test, float_reconstruction_error=1e-7):
...
             'all_32bit': mct.QuantizationConfig(mct.QuantizationErrorMethod.NOCLIPPING,
                                                 mct.QuantizationErrorMethod.NOCLIPPING,
-                                                32, 32, False, True, True,
+                                                False, True, True,
                                                 enable_weights_quantization=True,
                                                 enable_activation_quantization=True),
             'all_4bit': mct.QuantizationConfig(mct.QuantizationErrorMethod.NOCLIPPING,
                                                mct.QuantizationErrorMethod.NOCLIPPING,
-                                               4, 4, False, False, True,
+                                               False, False, True,
```
reuvenperetz, line 66: We can remove this method and refactor Pytorch tests to run only over different hw models (as all configs here are identical). But let's do it in a different PR. For now, it's ok. Just add a TODO please

tests/keras_tests/feature_networks_tests/feature_networks/nested_networks/nested_model_multiple_inputs_test.py
```diff
@@ -35,8 +38,13 @@ class NestedModelMultipleInputsTest(BaseKerasFeatureNetworkTest):
     def __init__(self, unit_test):
         super().__init__(unit_test)
 
+    def get_fw_hw_model(self):
+        hwm = generate_test_hw_model({'weights_n_bits': 16,
+                                      'activation_n_bits': 16})
+        return generate_fhw_model_keras(name="nested_multi_inputs_test", hardware_model=hwm)
+
     def get_quantization_config(self):
```
reuvenperetz, line 46: Since this model is heavily used in many tests, let's create a single one in tests helpers and reuse it.

reuvenperetz (changes requested):

PR #140, opened 2022-04-07
Activation mixed-precision configurations support 

ofirgo:
First step in implementation of activation mixed-precision.
Preparing the graph and node configuration to support activation mp.

### DIFF
--- a/model_compression_toolkit/common/graph/base_node.py
+++ b/model_compression_toolkit/common/graph/base_node.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 
 import copy
-from typing import Dict, Any, Tuple
+from typing import Dict, Any, Tuple, List
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


model_compression_toolkit/common/graph/base_node.py
```diff
@@ -290,3 +290,11 @@ def has_weights_to_quantize(self, fw_info):
...
                 return True
         return False
+
+    def get_total_output_params(self):
+        """
+        Calculates the output size of the node.
+
+        Returns: Output size.
+        """
+        return np.prod([x for x in self.output_shape if x is not None])
```
reuvenperetz, line 300: What if the node has multiple outputs?

model_compression_toolkit/common/quantization/node_quantization_config.py
```diff
@@ -298,7 +297,6 @@ def __eq__(self, other: Any) -> bool:
         return self.weights_quantization_fn == other.weights_quantization_fn and \
                self.weights_quantization_params_fn == other.weights_quantization_params_fn and \
                self.weights_channels_axis == other.weights_channels_axis and \
-               self.weights_quantization_params == other.weights_quantization_params and \
                self.weights_error_method == other.weights_error_method and \
```
reuvenperetz, line 300: Why these checks are removed?

reuvenperetz (changes requested):

model_compression_toolkit/common/quantization/node_quantization_config.py
ofirgo: resolved offline

PR #142, opened 2022-04-07
Fix bias correction for Keras Depthwise

reuvenperetz:
Add support for bias correction in Keras DepthwiseConv2d
when depth_multiplier is greater than 1.


haihabi:
Please link this to the open issue...

PR #152, opened 2022-04-17
Added print of approximated compression ratio

lior-dikstein:


model_compression_toolkit/common/post_training_quantization.py
```diff
@@ -137,6 +137,7 @@ def post_training_quantization(in_model: Any,
                         bit_widths_config)
 
     common.Logger.info(f'Approximated model size (in bytes): {tg.get_memory()}')
+    common.Logger.info(f'Approximated compression ratio: {round(tg.get_float_memory() / (tg.get_memory() + 1e-8), 3)}')
```
reuvenperetz, line 141: I think we need to take the float memory of the original graph (retrieved from the model reader) instead of tg.

reuvenperetz (changes requested):

model_compression_toolkit/common/post_training_quantization.py
haihabi: Why do you think so?

PR #157, opened 2022-04-19
Filtering-out irrelevant quantization config candidates 

ofirgo:
After marking activation, we check if there are nodes that have quantization candidates that are no longer relevant, due to disabling of weights/activation quantization, which can cause duplications in the candidates list.


model_compression_toolkit/common/post_training_quantization.py
```diff
@@ -436,6 +437,14 @@ def _prepare_model_for_quantization(graph: Graph,
     if tb_w is not None:
         tb_w.add_graph(transformed_graph, 'after_graph_marking')
 
+    ######################################
+    # Filter nodes' candidates
+    ######################################
+    filter_nodes_candidates(transformed_graph)
+
```
reuvenperetz, line 444: Please copy the graph and edit and return the copy.

model_compression_toolkit/common/quantization/filter_nodes_candidates.py
```diff
@@ -0,0 +1,74 @@
...
+
+    elif not node.is_weights_quantization_enabled():
+        # Remove candidates that have duplicated activation bitwidth for node with disabled weights quantization.
+        # Replacing the weights n_bits in the remained configurations with default value to prevent confusion.
+        seen_bitwidth = set()
+        filtered_candidates = [candidate for candidate in filtered_candidates if
+                               candidate.activation_quantization_cfg.activation_n_bits not in seen_bitwidth
+                               and not seen_bitwidth.add(candidate.activation_quantization_cfg.activation_n_bits)]
+        for c in filtered_candidates:
+            c.weights_quantization_cfg.weights_n_bits = DEFAULT_CANDIDATE_BITWIDTH
```
reuvenperetz, line 72: Filter not only by the bitwidth, but by the entire configuration

reuvenperetz (changes requested):

PR #173, opened 2022-05-01
KPI data API

ofirgo:
Added new API for retrieving useful information to be used for defining target KPI for mixed-precision optimization.

The API call returns a KPI object with the relevant data for weights and activations - sum of configurable weights layers' parameters, and size of max activation output tensor.

The new API is implemented and tested for both tf and pytorch frameworks.


model_compression_toolkit/keras/kpi_data_facade.py
```diff
@@ -0,0 +1,98 @@
...
+            Create a random dataset generator:
+
+            >>> import numpy as np
+            >>> def repr_datagen(): return [np.random.random((1,224,224,3))]
+
+            TODO: what is the correct way to explain about how to provide fw hw model? (do we need to give example for this if it is not mandatory?)
+            Import mct and call for KPI data calculation:
+            >>> import model_compression_toolkit as mct
+            >>> kpi_data_dict = keras_kpi_data(model, repr_datagen)
+
```
reuvenperetz, line 76: No need to explain here about HW model. Only mandatory arguments. But let's write in the function documentation about the default values. Something like: https://github.com/sony/model_optimization/blob/main/model_compression_toolkit/keras/quantization_facade.py#L117

reuvenperetz: And I would rename kpi_data_dict to kpi_data as this is not a dictionary.

reuvenperetz (changes requested):

model_compression_toolkit/keras/kpi_data_facade.py
ofirgo: It was a dict initially and I forgot to change the name, fixed

ofirgo: > No need to explain here about HW model. Only mandatory arguments. But let's write in the function documentation about the default values. Something like: https://github.com/sony/model_optimization/blob/main/model_compression_toolkit/keras/quantization_facade.py#L117

I already added a documentation about the default value for fw_info, which is the one that has default value in the quantization_facade (the link you added).
Do you mean we should add default to other arguments' documentation as well? which ones?

PR #187, opened 2022-05-02
add tf.math.argmax layer type to default hw model NoQuantization

ofirgo:


reuvenperetz:
I'll create a new one for this change due to conflicts

PR #206, opened 2022-05-09
added layers types for keras default tpc NoQuantization list

ofirgo:
Added TFOpLambda type layers to the list of NoQuantization in the default keras  tpc

### DIFF
--- a/model_compression_toolkit/tpc_models/keras_tp_models/keras_default.py
+++ b/model_compression_toolkit/tpc_models/keras_tp_models/keras_default.py
@@ -59,7 +59,10 @@ def generate_keras_default_tpc(name: str, tp_model: TargetPlatformModel):
                                                      MaxPooling2D,
                                                      tf.split,
                                                      tf.quantization.fake_quant_with_min_max_vars,
-                                                     tf.math.argmax])
+                                                     tf.math.argmax,
+                                                     tf.shape,
+                                                     tf.image.resize,
+                                                     tf.__operators__.getitem])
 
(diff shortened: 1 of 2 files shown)


model_compression_toolkit/tpc_models/keras_tp_models/keras_default.py
```diff
@@ -59,7 +59,10 @@ def generate_keras_default_tpc(name: str, tp_model: TargetPlatformModel):
                                                      MaxPooling2D,
                                                      tf.split,
                                                      tf.quantization.fake_quant_with_min_max_vars,
-                                                     tf.math.argmax])
+                                                     tf.math.argmax,
+                                                     tf.shape,
+                                                     tf.image.resize,
```
reuvenperetz, line 64: Why resize isn't quantized?

reuvenperetz (changes requested):

model_compression_toolkit/tpc_models/keras_tp_models/keras_default.py
ofirgo: It failed so I thought that it's like the others..
I was in the mind that it's like "Reshape", but if we should quantize it than we have some other issue with it that we need to figure out

ofirgo: Ok, I removed the resize from the list (currently, couldn't reproduce the issue that I thought was caused by the resize layer anyway)

PR #207, opened 2022-05-09
fix issues with pytorch mp wrapper and model builder

ofirgo:
Main issues was that the previous implementation didn't consider functional layers as configurable for mixed-precision quantization (didn't wrap them with the mp wrapper).
Also, minor issues with the mp_model setting for the relevant device (fixed).


model_compression_toolkit/pytorch/back2framework/model_builder.py
```diff
@@ -187,6 +188,22 @@ def forward(self,
                 outputs = outputs[0]
         return outputs
 
+    def get_op_func(self, n: BaseNode, configurable_nodes_names: List[str]) -> any:
+        """
```
reuvenperetz, line 192: Please fix the type hint

reuvenperetz (changes requested):

PR #209, opened 2022-05-09
Mixed precision metric interest points bound

ofirgo:
Allowing to bound the number of interest points that are being used for metric evaluation in MP search, by providing a factor argument in the MP quantization config.
The reduced set of interest points is composed of evenly spaced points from the original list of points.


model_compression_toolkit/common/mixed_precision/mixed_precision_quantization_config.py
```diff
@@ -28,7 +28,8 @@ def __init__(self,
                  compute_distance_fn: Callable = compute_mse,
                  distance_weighting_method: Callable = get_average_weights,
                  num_of_images: int = 32,
-                 configuration_overwrite: List[int] = None):
+                 configuration_overwrite: List[int] = None,
+                 num_interest_points_factor: float = 1.0):
         """
```
reuvenperetz, line 33: Please add to the documentation.

reuvenperetz (changes requested):

reuvenperetz (approved):

PR #214, opened 2022-05-12
Feature support TensorFlowOpLayer

ofirgo:
Add TensorFlowOpLayer type layers to graph as Functional nodes and add layer filters for several type of operators to the default keras TargetPlatformCapabilities (so they won't be quantized).
Created a new type of AttributeFilter - NestedAttributeFilter , to allow defining filters that match nested attributes.


haihabi:
let talk about this PR I don't like how the TPC code looks like.

PR #215, opened 2022-05-12
Remove reused nodes from activation configurable nodes list

ofirgo:


haihabi:
Again this is mixed up with the TFOpLamda PR. I am not sure what is the concrete change.

PR #216, opened 2022-05-12
Minor fix in BaseNode get_weights_by_keys method

ofirgo:
Return None if given None attribute name


haihabi:
This PR include also TFOPLambda change, If I understand correctly you want to add None output to this function?
I am not sure this is a good approach...

PR #225, opened 2022-05-15
Fix issue 223

reuvenperetz:


model_compression_toolkit/tpc_models/get_target_platform_capabilities.py
```diff
@@ -75,6 +75,9 @@ def get_target_platform_capabilities(fw_name: str,
     """
     assert fw_name in tpc_dict, f'Framework {fw_name} is not supported'
     supported_models_by_fw = tpc_dict.get(fw_name)
-    assert target_platform_name in supported_models_by_fw, f'Target platform model named {target_platform_name} is not' \
-                                              f' supported for framework {fw_name}'
+    assert supported_models_by_fw is not None, f'Supported models for {fw_name} is empty. Make sure the relevant ' \
```
haihabi, line 78: this can happen when the user puts a wrong framework name also... we need to validate that also.

reuvenperetz: This is checked in line 76

PR #227, opened 2022-05-16
Compute mixed-precision metric for best config only once

ofirgo:
Remove duplicate computation of the mixed precision distance metric when calculating the layer to metric mapping.
For each layer, when computing the distance metric value for its best candidate config, we actually compute the metric for the mp config with the max bitwidth for each layer, therefore, this is the same computation for each layer and we can run it just once.


tests/keras_tests/function_tests/test_lp_search_bitwidth.py
```diff
@@ -41,14 +41,19 @@
...
+    return 0
+
+
 class MockMixedPrecisionSearchManager:
     def __init__(self, layer_to_kpi_mapping):
         self.layer_to_bitwidth_mapping = {0: [0, 1, 2]}
         self.layer_to_kpi_mapping = layer_to_kpi_mapping
-        self.compute_metric_fn = lambda x, y: 0
+        self.compute_metric_fn = compute_metric_function
         self.min_kpi = {KPITarget.WEIGHTS: [[1], [1], [1]], KPITarget.ACTIVATION: [[1], [1], [1]]}  # minimal kpi in the tests layer_to_kpi_mapping
```
reuvenperetz, line 53: Why not leave it as lambda?

ofirgo: Because the second argument to the function is optional (has default value 'None'), and it now actually use it at some run path after those changes.

ofirgo: Ok, I found out just now that lambda function can also get default value for arguments, so I'll change it back and add the default value to the lambda.

reuvenperetz (approved):

PR #241, opened 2022-05-19
Mixed-precision result visualizer

ofirgo:
Adding mixed-precision result visualization.
Translating the final bitwidth config (candidate indices) to actual weights and activation quantization bitwidth, printing to logger and adding figures to tensorboard.


model_compression_toolkit/common/graph/base_graph.py
```diff
@@ -627,3 +627,23 @@ def get_max_candidates_config(self) -> List[int]:
             f"A maximal config candidate must be defined, but some node have multiple potential maximal candidates"
 
         return [lst[0] for lst in max_cfg_candidates]
+
+    def get_final_weights_config(self):
+        """
```
reuvenperetz, line 632: Please add type hints

model_compression_toolkit/common/graph/base_graph.py
```diff
@@ -627,3 +627,23 @@ def get_max_candidates_config(self) -> List[int]:
...
+        Gets the final number of bits for quantization of each weights' configurable layer.
+
+        Returns: A list of pairs of (node type, nod's weights quantization bitwidth).
+
+        """
+        sorted_conf_weights = self.get_sorted_weights_configurable_nodes()
+        return [(n, n.final_weights_quantization_cfg.weights_n_bits) for n in sorted_conf_weights]
+
+    def get_final_activation_config(self):
+        """
```
reuvenperetz, line 642: Please add type hints

model_compression_toolkit/common/visualization/final_config_visualizer.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Args:
+        node: A graph node representing a model's layer.
+
+    Returns: The name of the layer to appear in the visualization.
+
+    """
+
+    return node.type.__name__
+
```
reuvenperetz, line 34: Does it work for functional nodes as well? (does tf.add has it?)

model_compression_toolkit/common/visualization/final_config_visualizer.py
```diff
@@ -0,0 +1,147 @@
...
+# ==============================================================================
+from typing import Tuple, List
+from matplotlib import pyplot as plt
+from matplotlib.figure import Figure
+
+from model_compression_toolkit.common import Graph
+
+
+def get_layer_represent_name(node):
+    """
```
reuvenperetz, line 23: Type hints

reuvenperetz (approved):

model_compression_toolkit/common/visualization/final_config_visualizer.py
ofirgo: I'm not sure, I can check, but anyway we only apply it to nodes that have kernels that we want to quantize (we present the name only for the weights configurable nodes)

reuvenperetz: If it's relevant only for weights configurable, so let's rename the function name to indicate it. Also, we need to check for example the case of tf.nn.conv2d (which is functional)

PR #263, opened 2022-06-02
Modified per-layer distance function for MP distance metric

ofirgo:
Modifications to distance function for mp metric computation (keras only):

- Specific distance metric for some nodes' types (defaults in normalized MSE)
- Option to use given distance function for all nodes as before (with given argument)
- Several nodes' types added to interest points set
- Required modification to sensitivity evaluation metric
- Added KL divergence function for similarity analysis


model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation_manager.py
```diff
@@ -137,34 +154,32 @@ def build_distance_metrix(self):
...
+    interest_points_nodes = bound_num_interest_points(ip_nodes, num_ip_factor)
 
-    # We add the last configurable node to the list of interest points, since we need it to be able to include all
-    # configurable nodes in the sensitivity metric computation
+    # We add last layer in model to interest points,
+    # in order to consider the model's output in the distance metric computation (and also to make sure
+    # all configurable layers are included in the configured mp model for metric computation purposes)
+    last_layer_node = sorted_nodes[-1]
+    interest_points = interest_points_nodes if last_layer_node in interest_points_nodes \
+        else interest_points_nodes + [last_layer_node]
```
reuvenperetz, line 182: I think we should take all outputs instead of just the last node in the sorted nodes

model_compression_toolkit/core/keras/keras_implementation.py
```diff
@@ -319,3 +326,21 @@ def get_node_prior_info(self,
...
         return create_node_prior_info(node=node,
                                       fw_info=fw_info, graph=graph)
+
+    def count_node_for_mixed_precision_interest_points(self, node: BaseNode) -> bool:
+        """
+        Returns whether a given node in considered as a potential interest point for mp metric computation purposes.
+        Args:
+            node: Node to indicate whether it needs to be part of the interest points set.
+        Returns: True if the node should be considered an interest point, False otherwise.
+        """
```
reuvenperetz, line 336: Why do we need this function instead of using the configurable nodes?

reuvenperetz (changes requested):

model_compression_toolkit/core/keras/reader/node_builder.py
```diff
@@ -129,3 +138,25 @@ def __is_functional_inputs_a_list(op_call_args: Any) -> bool:
...
+
+    if layer_class == Activation:
+        node_type_name = layer_config[ACTIVATION]
+        if node_type_name == SOFTMAX or node_type_name == SIGMOID:
+            return compute_kl_divergence
+    elif layer_class == tf.nn.softmax or layer_class == tf.nn.sigmoid:
+        return compute_kl_divergence
+    elif layer_class == Dense:
+        return compute_cs
+    return lambda x, y: compute_mse(x, y, norm=True, norm_eps=1e-8)
```
reuvenperetz, line 162: Maybe we should move it to fw implementation and create a function to get interest-points (in fw implementation as well). The get_distance_fn can then use the interest points function to return the correct distance_fn (or None in case it's not an interest point).

reuvenperetz (changes requested):

model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation_manager.py
```diff
@@ -83,6 +87,17 @@ def _get_baseline_tensors(self):
...
+        """
+        If a specific distance function was provided, then it is returned to be used for distance metric computation
+        for all nodes. Otherwise, the specific distance function configured for each specific node is returned.
+        Returns: A distance function between to tensors, to be used for metric computation.
+        """
+        if self.quant_config.compute_distance_fn is not None:
+            return lambda x, y, idx: self.quant_config.compute_distance_fn(x, y)
+        else:
+            return lambda x, y, idx: self.interest_points[idx].distance_fn(x, y)
+
```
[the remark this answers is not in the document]
ofirgo: Changed the structure of this implementation completely, so this method was removed

PR #269, opened 2022-06-09
Pytorch remove asserts from graph

lapid92:
Remove from Pytorch graph reader nodes without output (like assert)

### DIFF
--- a/model_compression_toolkit/core/pytorch/reader/reader.py
+++ b/model_compression_toolkit/core/pytorch/reader/reader.py
@@ -91,12 +91,48 @@ def fx_graph_module_generation(pytorch_model: torch.nn.Module,
     return symbolic_traced
 
 
+def remove_broken_nodes_from_graph(graph):
+    """
+    Remove all "broken" nodes (nodes without output).
+
+    Args:
+        graph: Networkx MultiDiGraph representing the Pytorch model.
+
+    Returns:
+        Networkx MultiDiGraph representing the Pytorch model after removing nodes without output (like assert).
+
+    """
+    output_nodes = [n.node for n in graph.output_nodes]
+    # make a list of all the nodes with no output
+    nodes_list = [n for n, d in graph.out_degree if d == 0 and n not in output_nodes]
+    nodes_to_remove = []
+    for node in nodes_list:
+        nodes_to_remove.append(node)
+        # iterate over all the relevant nodes
+        while len(nodes_to_remove) >> 0:
+            for node_to_remove in nodes_to_remove:
+                parent_nodes = [edge.source_node for edge in graph.incoming_edges(node_to_remove)]
+                # check all the parent nodes
+                for parent_node in parent_nodes:
+                    # if the parent node connected only to the "broken" node, we'll add the parent node to
+                    # the relevant nodes list
+                    # if the parent node connected to other nodes, we'll only remove the edge
+                    if graph.out_degree(parent_node) == 1:
+                        nodes_to_remove.append(parent_node)
+                    graph.remove_edge(parent_node, node_to_remove)
+                # remove the current node
+                graph.remove_node(node_to_remove)
+                nodes_to_remove.remove(node_to_remove)
+    return graph
+
+
 def model_reader(model: torch.nn.Module,
                  representative_data_gen: Callable,
                  to_numpy: Callable,
                  to_tensor: Callable) -> Graph:
     """
-    Reads a Pytorch model, converts it to an FX Graph using the fx toolkit, then builds a base graph representing the fx graph.
+    Reads a Pytorch model, converts it to an FX Graph using the fx toolkit, then builds a base graph representing the
+    fx graph and filter "broken nodes" (like assert).
     Args:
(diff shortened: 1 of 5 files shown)


model_compression_toolkit/core/pytorch/reader/reader.py
```diff
@@ -91,12 +91,48 @@ def fx_graph_module_generation(pytorch_model: torch.nn.Module,
...
+        nodes_to_remove.append(node)
+        # iterate over all the relevant nodes
+        while len(nodes_to_remove) >> 0:
+            for node_to_remove in nodes_to_remove:
+                parent_nodes = [edge.source_node for edge in graph.incoming_edges(node_to_remove)]
+                # check all the parent nodes
+                for parent_node in parent_nodes:
+                    # if the parent node connected only to the "broken" node, we'll add the parent node to
+                    # the relevant nodes list
+                    # if the parent node connected to other nodes, we'll only remove the edge
```
lior-dikstein, line 119: (typo) # if the parent node **is** connected only to the "broken" node, we'll add the parent node to
                    # the relevant nodes list.
                    # if the parent node **is** connected to other nodes, we'll only remove the edge

model_compression_toolkit/core/pytorch/reader/reader.py
```diff
@@ -91,12 +91,48 @@ def fx_graph_module_generation(pytorch_model: torch.nn.Module,
...
+
+    """
+    output_nodes = [n.node for n in graph.output_nodes]
+    # make a list of all the nodes with no output
+    nodes_list = [n for n, d in graph.out_degree if d == 0 and n not in output_nodes]
+    nodes_to_remove = []
+    for node in nodes_list:
+        nodes_to_remove.append(node)
+        # iterate over all the relevant nodes
+        while len(nodes_to_remove) >> 0:
```
lior-dikstein, line 112: what is " >>"? Is it a typo or is it ok? I am not familiar with this

model_compression_toolkit/core/pytorch/reader/reader.py
```diff
@@ -91,12 +91,48 @@ def fx_graph_module_generation(pytorch_model: torch.nn.Module,
...
+
+
 def model_reader(model: torch.nn.Module,
                  representative_data_gen: Callable,
                  to_numpy: Callable,
                  to_tensor: Callable) -> Graph:
     """
-    Reads a Pytorch model, converts it to an FX Graph using the fx toolkit, then builds a base graph representing the fx graph.
+    Reads a Pytorch model, converts it to an FX Graph using the fx toolkit, then builds a base graph representing the
+    fx graph and filter "broken nodes" (like assert).
```
lior-dikstein, line 135: Reads a Pytorch model and converts it to an FX Graph using the fx toolkit. Then, builds a base graph representing the
    fx graph. Finally, we filter "broken nodes" (nodes without outputs, for example: "assert").

tests/pytorch_tests/model_tests/feature_models/add_net_assert_test.py
```diff
@@ -0,0 +1,52 @@
...
+import torch
+from torch import _assert
+from tests.pytorch_tests.model_tests.base_pytorch_test import BasePytorchTest
+
+"""
+This tests checks that the assert operation is being removed from the graph during quantization.
+"""
+
+
+class AddAssertNet(torch.nn.Module):
```
lior-dikstein, line 24: Please change the class name to "AssertNet", and the file name to "remove_assert_test.py"

tests/pytorch_tests/model_tests/feature_models/add_net_assert_test.py
```diff
@@ -0,0 +1,52 @@
...
+        _assert(x.shape == y.shape, "Inputs should have the same shape")
+        x1 = self.conv1(x)
+        assert(x.ndim == 4, f'expected 4D input (got {x.ndim}D input)')
+        x2 = x1 + 3
+        _assert(x2.shape == x1.shape, "Tensors should have the same shape")
+        y = self.conv2(y)
+        return x2 - y, y - x2, x2 + y
+
+
+class AddAssertNetTest(BasePytorchTest):
```
lior-dikstein, line 40: Change class name, same as above

tests/pytorch_tests/model_tests/feature_models/add_net_assert_test.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    def __init__(self, unit_test):
+        super().__init__(unit_test)
+
+    def create_inputs_shape(self):
+        return [[self.val_batch_size, 3, 32, 32], [self.val_batch_size, 3, 32, 32]]
+
+    def create_feature_network(self, input_shape):
+        return AddAssertNet()
```
lior-dikstein, line 52: How do you check that the "assert" nodes have been removed?
I would also add a test with a node that is just "floating" (i.e. doesn't have an output or doesn't contribute to the network output) and see that it is removed while other nodes remain.

lior-dikstein (changes requested):

PR #274, opened 2022-06-19
refactor GPTQ

elad-c:
Move all GPTQ files to GPTQ folder (keras only)


model_compression_toolkit/core/keras/quantization_facade.py
```diff
@@ -18,12 +18,12 @@
 from model_compression_toolkit.core import common
 from model_compression_toolkit.core.common import Logger
 from model_compression_toolkit.core.common.constants import TENSORFLOW
-from model_compression_toolkit.core.common.gptq.gptq_config import GradientPTQConfig
+from model_compression_toolkit.gptq.common.gptq_config import GradientPTQConfig
 from model_compression_toolkit.core.common.mixed_precision.kpi import KPI
 from model_compression_toolkit.core.common.framework_info import FrameworkInfo
 from model_compression_toolkit.core.common.network_editors.actions import EditRule
 from model_compression_toolkit.core.common.mixed_precision.mixed_precision_quantization_config import \
-    MixedPrecisionQuantizationConfig, DEFAULT_MIXEDPRECISION_CONFIG, MixedPrecisionQuantizationConfigV2
```
ofirgo, line 26: Not necessarily related to this PR but why aren't we using version 2 (MixedPrecisionQuantizationConfigV2) in the facade? Aren't we going to switch eventually to version 2 (and make it the only version)?

ofirgo (approved):
Just one minor non-blocking comment

model_compression_toolkit/core/keras/quantization_facade.py
elad-c: because this is the old (depricated) API. It will be removed in the future.

PR #275, opened 2022-06-19
Fix pytorch builder for getattr and add unbind to non-quant list

eladc-git:
1. Fix getattr operation in pytorch graph builder
2. Add unbind (pytorch) to non-quantization TPC list

### DIFF
--- a/model_compression_toolkit/core/pytorch/reader/graph_builders.py
+++ b/model_compression_toolkit/core/pytorch/reader/graph_builders.py
@@ -86,6 +86,11 @@ def nodes_builder(model: GraphModule,
                 framework_attr[BIAS] = False if node_module.bias is None else True
         elif node.op == CALL_FUNCTION:
             node_type = node.target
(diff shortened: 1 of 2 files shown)


model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -107,6 +107,10 @@ def nodes_builder(model: GraphModule,
         else:
             raise Exception(f'Unknown node type: {node.name}')
 
+        # Disable activation
+        if node.target in [UNBIND,getattr]:
```
haihabi, line 111: Please don't merge it @ofirgo @lior-dikstein  this looks like a patch

eladc-git: Please review again. If still it is not Ok let's talk.

haihabi (approved):

lior-dikstein (approved):

PR #282, opened 2022-06-29
Remove node from kwargs pytorch

lapid92:
Remove requires grad flag from pytorch model
Add bn function test


haihabi:
Not relevant...

PR #287, opened 2022-07-05
PyTorch support for gradient-based weights for mixed-precision metric

ofirgo:
* Added support to gradient-based weights for the distance metric computation in Pytorch framework.
* Default value for alpha parameter (tuning between output layers and regular layers) is changed to 0.3 (based on experimental study).
* Default weights for metric computation remained uniform (gradient-based weights option needs to be activated by an argument to mixed-precision quantization config).
* Minor fix in Keras gradient model tests TPC imports.

### DIFF
--- a/model_compression_toolkit/core/pytorch/back2framework/model_gradients.py
+++ b/model_compression_toolkit/core/pytorch/back2framework/model_gradients.py
@@ -65,7 +65,6 @@ def run_operation(n: BaseNode,
         n: The corresponding node of the layer it runs.
         input_tensors: List of Pytorch tensors that are the layer's inputs.
         op_func: Module/functional to apply to the input tensors.
-        mode: model quantiztion mode from ModelBuilderMode
     Returns:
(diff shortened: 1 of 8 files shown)


model_compression_toolkit/core/pytorch/back2framework/model_gradients.py
```diff
@@ -0,0 +1,256 @@
...
+
+            op_func = n.type if isinstance(n, FunctionalNode) else getattr(self, n.name)
+            out_tensors_of_n = run_operation(n,  # Run node operation and fetch outputs
+                                             input_tensors,
+                                             op_func=op_func)
+
+            if isinstance(out_tensors_of_n, list):
+                output_t = []
+                for t in out_tensors_of_n:
+                    t.retain_grad()
```
reuvenperetz, line 153: 1. Why we don't check here if the tensor requires_grad?
2. Shouldn't we use gradients only for interest points?

PR #290, opened 2022-07-06
Activation LUT Kmean Quantizer (TF)

ofirgo:
Implemented activation lookup table activation quantizer for TensorFlow.
- Added parameters selection for histogram with weighted kmeans (for finding clusters)
- Implemented custom Keras layer LUTFakeQuant for non-uniform quantization of activation.

### DIFF
--- a/model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
+++ b/model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
@@ -10,7 +10,7 @@ from model_compression_toolkit.core.common.constants import SIGNED, CLUSTER_CENT
 
 
 def activation_lut_kmean_quantizer(activation_n_bits: int,
-                                   quantization_params: dict) -> Callable:
+                                   quantization_params: Dict[str, np.ndarray]) -> Callable:
     """
     Builds a LUT quantizer for layer's activation using the provided params (threshold and clusters).
     It initiates a fake custom LUT layer that provides the quantizer function.
@@ -49,7 +49,7 @@ class LUTFakeQuant(Layer):
         """
         super(LUTFakeQuant, self).build(input_shape)
 
-    def call(self, input_data: Any, **kwargs) -> TFReference:
+    def call(self, input_data: TFReference, **kwargs) -> TFReference:
         """
 
         Args:
@@ -65,7 +65,7 @@ class LUTFakeQuant(Layer):
         _quant_output = self.lut_kmeans_quantizer(input_data)
         return _quant_output
 
-    def lut_kmeans_quantizer(self, tensor_data: Any) -> TFReference:
+    def lut_kmeans_quantizer(self, tensor_data: TFReference) -> TFReference:
         """
         Quantize a tensor using a non-uniform quantization based on the pre-defined kmeans clusters.
         1. Scales tensor_data with the threshold into 8-bit quantization range.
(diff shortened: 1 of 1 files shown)


model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
```diff
@@ -51,10 +50,10 @@ def lut_kmeans_tensor(tensor_data: np.ndarray,
 
     Returns:
         A dictionary containing the cluster assignments according to the k-means algorithm,
-        the scales per channel and the multiplier num bits.
+        the thresholds per channel and the multiplier num bits.
     """
-    if n_bits > MULTIPLIER_N_BITS:
-        raise Exception(f'Look-Up-Table bit configuration has {n_bits} bits. It must be less or equal to {MULTIPLIER_N_BITS}')
+    if n_bits >= MULTIPLIER_N_BITS:
```
reuvenperetz, line 55: Why the equal was added?

model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+
+        Args:
+            input_data: A Keras input tensor.
+            **kwargs: Optional arguments' dictionary.
+
+        Returns: KerasTensor after applying a non-uniform fake quantization.
+
+        """
+        if self.activation_is_signed is None or self.cluster_centers is None or self.threshold is None:
```
reuvenperetz, line 62: I think we should assert it in the init method, instead of returning None. What do you think?

model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
```diff
@@ -0,0 +1,118 @@
...
+
+        quant_tensor = (centers / (2 ** (MULTIPLIER_N_BITS - int(self.activation_is_signed)))) * self.threshold
+
+        return quant_tensor
+
+    def int_quantization_with_threshold(self,
+                                        data: TFReference,
+                                        n_bits: int,
+                                        eps: float = EPS) -> TFReference:
+        """
```
reuvenperetz, line 97: Why don't we use int_quantization_with_threshold that is in quantizers_helpers?

reuvenperetz (changes requested):

model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
ofirgo: I discussed it with Hai and we came to an understanding that 8-bit lut quantization is not practical (and sometimes can collapse the program due to memory issues).

model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
ofirgo: This is the behavior in all the quantizers (if the parameters values are None we return None).
If we change it here we need to change it in all the other quantizers.

model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
ofirgo: Because it is works on np arrays and we need an implementation for tf.Tensors

model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
reuvenperetz: Ok. In general I think it's better to assert, but let's leave it for a different PR

model_compression_toolkit/core/keras/quantizer/lut_fake_quant.py
reuvenperetz: Ok. So I just think maybe it's better to create a similar file (like quantizers utils) for TF quantizers, as this it shouldn't be coupled to LUT specifically. What do you think?

PR #299, opened 2022-07-17
Add substitution: LayerNorm --> LayerNorm+BatchNorm

elad-c:
A substitution to move LayerNorm weights (gamma and beta) to an appended BatchNorm.
The substitution adds a BatchNorm after the LayerNorm with gamma & beta copied from LayerNorm.
Then, LayerNorm gamma & beta are set to 1 & 0 respectively.
BatchNorm epsilon, moving_avg & moving_var are set to 0, 0 & 1 respectively.


model_compression_toolkit/core/keras/graph_substitutions/substitutions/layer_norm_decomposition.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    def __init__(self):
+        """
+        Matches LayerNorm node.
+        """
+        super().__init__(matcher_instance=NodeOperationMatcher(LayerNormalization))
+
+    @staticmethod
+    def _get_weight_by_name(mha_node, w_str):
```
haihabi, line 44: 1. Comment 
2. Rename mha_node to something else

tests/keras_tests/feature_networks_tests/feature_networks/layer_norm_substitution_test.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2021 Sony Semiconductors Israel, Inc. All rights reserved.
```
haihabi, line 1: This is new file and the year is 2022

PR #301, opened 2022-07-19
Jacobian trace approximation for weighted average distance metric.

ofirgo:
Implementing Jacobian trace approximation to be used for distance meאric weighted average.
This new method replaces the Hessian trace approximation as the model gradient feature for computing gradient-based weights.

### DIFF
--- a/model_compression_toolkit/core/common/framework_implementation.py
+++ b/model_compression_toolkit/core/common/framework_implementation.py
@@ -367,9 +367,10 @@ class FrameworkImplementation(ABC):
                    interest_points: List[BaseNode],
                    output_list: List[BaseNode],
                    all_outputs_indices: List[int],
-                   alpha: float = 0.3) -> List[float]:
+                   alpha: float = 0.3,
+                   n_iter: int = 50) -> List[float]:
         """
-        Calls a framework specific model gradient calculation function, which computes the gradients of the model's
+        Calls a framework specific model gradient calculation function, which computes the jacobian of the model's
         outputs with respect to the feature maps of the set of given interest points.
 
         Args:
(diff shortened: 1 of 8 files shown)


model_compression_toolkit/core/common/framework_implementation.py
```diff
@@ -382,8 +383,9 @@ def model_grad(self,
             alpha: A tuning parameter to allow calibration between the contribution of the output feature maps returned
                 weights and the other feature maps weights (since the gradient of the output layers does not provide a
                 compatible weight for the distance metric computation).
+            n_iter: The number of random iterations to calculate the approximated jacobian trace for each interest point.
 
-        Returns: A list of normalized gradients to be considered as the relevancy that each interest
+        Returns: A list of normalized jacobians traces to be considered as the relevancy that each interest
```
haihabi, line 388: We are computing the trace(jacobians jacobians^T) and not the jacobians traces

model_compression_toolkit/core/pytorch/back2framework/model_gradients.py
```diff
@@ -105,7 +106,8 @@ class PytorchModelGradients(torch.nn.Module):
     """
     def __init__(self,
                  graph_float: common.Graph,
-                 interest_points: List[BaseNode]):
+                 interest_points: List[BaseNode],
+                 output_list: List[BaseNode]):
         """
         Construct a PytorchModelGradients model.
```
haihabi, line 113: Missing argument in comments

model_compression_toolkit/core/common/framework_implementation.py
ofirgo: Is the term "jacobian-based weights" sounds ok? It's not possible to mention  trace(jacobians jacobians^T) everywhere in the comments (it's not just here, and it wouldn't be clear)

PR #317, opened 2022-08-03
Incorporating weighted loss for GPTQ

ofirgo:
Added option to use jacobian-based weights to compute the GPTQ training loss.
The weights are computed using the framework's "model_gradients" tool (which is also used for mixed-precision distance metric weights).
In addition, several updates to the model gradients computation:
- Added sqrt to the final computed mean squared jacobian approximation.
- Added an option to compute non-normalized weights (archives improves results).


model_compression_toolkit/gptq/keras/gptq_loss.py
```diff
@@ -39,7 +39,7 @@ def multiple_tensors_mse_loss(y_list: List[tf.Tensor],
                               flp_w_list: List[List[tf.Tensor]],
                               act_bn_mean: List,
                               act_bn_std: List,
-                              ) -> tf.Tensor:
+                              loss_weights: List[float] = None) -> tf.Tensor:
```
haihabi, line 42: Missing comment

model_compression_toolkit/gptq/keras/gptq_training.py
```diff
@@ -91,6 +93,8 @@ def __init__(self,
         else:
             self.input_scale = self.gptq_user_info.input_scale
 
+        self.weights_for_average_loss = self._compute_jacobian_based_weights(gptq_config, representative_data_gen)
```
reuvenperetz, line 96: don't pass gptq_config. use it from self

reuvenperetz (changes requested):

reuvenperetz (approved):

PR #319, opened 2022-08-03
Support DW-Conv2D in QAT when quantizing per weights per channel

elad-c:
Fix bug with weight quantization per output channel when channel axis isn't the last axis


haihabi:
Please add some unit testing for this case.

reuvenperetz (approved):

elad-c:
Do not merge before **fix_weight_quantization** is merged

PR #322, opened 2022-08-04
fix weight quantization

elad-c:
Weight threshold calculation should always be assuming signed quantization

### DIFF
--- a/model_compression_toolkit/core/common/quantization/quantization_params_generation/power_of_two_selection.py
+++ b/model_compression_toolkit/core/common/quantization/quantization_params_generation/power_of_two_selection.py
@@ -23,6 +23,7 @@ from model_compression_toolkit.core.common.quantization.quantization_params_gene
     get_threshold_selection_tensor_error_function, get_threshold_selection_histogram_error_function
 from model_compression_toolkit.core.common.target_platform import QuantizationMethod
 
+
 def power_of_two_selection_tensor(tensor_data: np.ndarray,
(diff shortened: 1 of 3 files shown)


model_compression_toolkit/core/common/quantization/quantization_params_generation/power_of_two_selection.py
```diff
@@ -54,7 +55,7 @@ def power_of_two_selection_tensor(tensor_data: np.ndarray,
         tensor_max = get_tensor_max(tensor_data, per_channel, channel_axis)
         threshold = max_power_of_two(tensor_max, min_threshold)
     else:
-        signed = np.any(tensor_data < 0)
+        signed = True  # weights are always signed
```
haihabi, line 58: Please also check symmetric quantization

haihabi:
Add a unit tests that checks the quantizer are always signed.

reuvenperetz (approved):

model_compression_toolkit/core/common/quantization/quantization_params_generation/power_of_two_selection.py
elad-c: isn't power of 2 always symmetric?

reuvenperetz:
> Add a unit tests that checks the quantizer are always signed.

Maybe we should just add a TODO for now until an initial exporter version is out, so things like that can be easily checked.

PR #330, opened 2022-08-09
Update README and workflows to run tests by framework version

reuvenperetz:
Update README with new MCT changes (follows new block diagram).
Change github actions to test MCT per framework per version and use it in new README.


tests/keras_tests/models_tests/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2022 Sony Semiconductors Israel, Inc. All rights reserved.
```
haihabi, line 1: Fix licenses remove s

haihabi (approved):

PR #334, opened 2022-08-11
BOPs KPI computation

ofirgo:
This is the second PR in the set of PRs which add BOPs KPI to mixed-precision. Need to be merged AFTER the first PR.

### DIFF
--- a/model_compression_toolkit/core/common/graph/virtual_activation_weights_node.py
+++ b/model_compression_toolkit/core/common/graph/virtual_activation_weights_node.py
@@ -17,13 +17,12 @@ from typing import Dict, Any, Tuple
 
 from model_compression_toolkit import FrameworkInfo
 from model_compression_toolkit.core.common.constants import VIRTUAL_ACTIVATION_WEIGHTS_NODE_PREFIX, \
-    VIRTUAL_WEIGHTS_SUFFIX, DEFAULT_CANDIDATE_BITWIDTH, VIRTUAL_ACTIVATION_SUFFIX
+    VIRTUAL_WEIGHTS_SUFFIX, DEFAULT_CANDIDATE_BITWIDTH, VIRTUAL_ACTIVATION_SUFFIX, ACTIVATION, LINEAR
 from model_compression_toolkit.core.common.graph.base_node import BaseNode
 import numpy as np
 
 from model_compression_toolkit.core.common.quantization.candidate_node_quantization_config import \
     CandidateNodeQuantizationConfig
-from model_compression_toolkit.core.keras.constants import ACTIVATION, LINEAR
 
 
 class VirtualSplitNode(BaseNode):
@@ -31,7 +30,7 @@ class VirtualSplitNode(BaseNode):
     A class that represents a node that was split from a kernel node (node with weights).
     """
 
-    def __init__(self, origin_node):
+    def __init__(self, origin_node: BaseNode):
         """
         Init a VirtualSplitNode object.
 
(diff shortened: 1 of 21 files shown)


model_compression_toolkit/core/common/mixed_precision/kpi_tools/kpi_methods.py
```diff
@@ -133,6 +159,156 @@ def total_weights_activation_kpi(mp_cfg: List[int], graph: Graph, fw_info: Frame
...
+    if not set_constraints:
+        return _bops_kpi(mp_cfg,
+                         graph,
+                         fw_info,
+                         fw_impl)
+
+    virtual_bops_nodes = [n for n in graph.get_topo_sorted_nodes() if isinstance(n, VirtualActivationWeightsNode)]
+
+    mp_nodes = graph.get_configurable_sorted_nodes_names()
+    bops = [n.get_bops_count(fw_impl, fw_info, candidate_idx=_get_node_cfg_idx(n, mp_cfg, mp_nodes)) for n in virtual_bops_nodes]
```
reuvenperetz, line 193: Shouldn't we compute BOPs for the interest points and not configurable nodes?

model_compression_toolkit/core/common/graph/virtual_activation_weights_node.py
```diff
@@ -183,3 +183,21 @@ def __init__(self,
...
+            candidate_idx: The index of the node's quantization candidate configuration.
+
+        Returns: The BOPS count of the composed node.
+
+        """
+        node_mac = fw_impl.get_node_mac_operations(self.original_weights_node, fw_info)
+        node_bops = self.candidates_quantization_cfg[candidate_idx].weights_quantization_cfg.weights_n_bits * \
+                    self.candidates_quantization_cfg[candidate_idx].activation_quantization_cfg.activation_n_bits * \
+                    node_mac
+        return node_bops
```
reuvenperetz, line 203: Why not checking if quantization is enabled (if not use FLOAT_BITWIDTH)?

reuvenperetz (changes requested):

model_compression_toolkit/core/keras/keras_implementation.py
```diff
@@ -484,3 +484,37 @@ def is_node_compatible_for_mp_metric_outputs(self,
...
+            return np.prod([x for x in output_shape if x is not None]) * \
+                   input_shape[input_channel_axis] * \
+                   (kernel_shape[0] * kernel_shape[1])
+        elif node.type is DepthwiseConv2D:
+            # Depth * (W_out * H_out) * C_in * (W_kernel * H_kernel)
+            return node.framework_attr.get(DEPTH_MULTIPLIER) * \
+                   np.prod([x for x in output_shape if x is not None]) / output_shape[output_channel_axis] * \
+                   input_shape[input_channel_axis] * \
+                   (kernel_shape[0] * kernel_shape[1])
+        elif node.type is Dense:
```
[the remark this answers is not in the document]
ofirgo: Isn't it the case for an FC layer? (I'm not sure but isn't the kernel size is [Input_size x Output_size] )

model_compression_toolkit/core/common/mixed_precision/kpi_tools/kpi_methods.py
ofirgo: No, interest points are only used for distance metric computation.
We need to compute the KPI for each node that contributes to it.

model_compression_toolkit/core/common/graph/virtual_activation_weights_node.py
ofirgo: Not sure to which part of the code this comment is referring

reuvenperetz (approved):

PR #339, opened 2022-08-16
Move some GPTQ functions to common

eladc-git:


model_compression_toolkit/gptq/common/gptq_training.py
```diff
@@ -63,6 +65,104 @@ def __init__(self,
 
         self.fxp_model, self.gptq_user_info = self.build_gptq_model()
 
+    def get_optimizer_with_param(self,
+                                 flattened_trainable_weights,
+                                 flattened_bias_weights,
+                                 trainable_quantization_parameters,
+                                 temperature_weights):
```
reuvenperetz, line 72: Please add comments and type hints

reuvenperetz (changes requested):

model_compression_toolkit/gptq/common/gptq_training.py
eladc-git: done

reuvenperetz (approved):

haihabi:
I am not sure about this PR since there is a significant difference between Keras and PyTorch optimizers.
Let's talk about it. Maybe we need to change Keras API.

eladc-git:
> I am not sure about this PR since there is a significant difference between Keras and PyTorch optimizers. Let's talk about it. Maybe we need to change Keras API.

This code is only optimizer object and weights holder so it is not supposed to be framework related.

PR #348, opened 2022-08-21
Fix optimizer import from Keras to support 2.9

reuvenperetz:


model_compression_toolkit/gptq/keras/quantization_facade.py
```diff
@@ -39,7 +39,8 @@
     from model_compression_toolkit.core.keras.keras_model_validation import KerasModelValidation
     from tensorflow.keras.models import Model
     from model_compression_toolkit.gptq.keras.gptq_loss import multiple_tensors_mse_loss
-    from keras.optimizer_v2.optimizer_v2 import OptimizerV2
+    from keras.optimizers.optimizer_v2.optimizer_v2 import OptimizerV2
```
haihabi, line 42: We don't need If else depending on the version ?

haihabi (approved):

PR #378, opened 2022-09-20
Move graph's fusions tracking to dedicated fused_nodes field

ofirgo:
Currently we save the fusion nodes lists in the graph's user_info.
This prevents us from using the record of fused nodes in other places in the code (since it is bad practice to user the "user_info" for features implementation).

This change only moves the record of fused nodes to a different variable in the graph object.

Note that currently the record remains also in the use info for tests compatibility (a TODO comment was added to remind that we should change the tests and remove it from user info).


eladc-git (approved):

ofirgo:
> Looking good. Why not changing the tests in this PR instead TODO?

Fixed the tests and removed the TODO comment

eladc-git:
> > Looking good. Why not changing the tests in this PR instead TODO?
> 
> Fixed the tests and removed the TODO comment

What about Pytorch tests?

ofirgo:
> > > Looking good. Why not changing the tests in this PR instead TODO?
> > 
> > 
> > Fixed the tests and removed the TODO comment
> 
> What about Pytorch tests?

Missed it, fixed now

PR #387, opened 2022-09-28
Add version field to TPC and new DEFAULT TPC v4

reuvenperetz:
This commit adds a field to the TargetPlatformCapabilities for version in order to assert the correct requested version.
Also, a new default TPC was added named v4 which uses SYMMETRIC quantizer for weights quantization.


.github/workflows/run_tests_suite_all_latest_frameworks.yml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 3000
```
ofirgo, line 13: Is this the tests suit time limit? isn't 30,000 minutes too much?

tests/keras_tests/test_keras_tp_model.py
```diff
@@ -235,12 +235,24 @@ def rep_data():
...
 
     def test_get_keras_supported_version(self):
-        mct.get_target_platform_capabilities(TENSORFLOW, DEFAULT_TP_MODEL)
-        mct.get_target_platform_capabilities(TENSORFLOW, DEFAULT_TP_MODEL, "v1")
-        mct.get_target_platform_capabilities(TENSORFLOW, DEFAULT_TP_MODEL, "v2")
-        mct.get_target_platform_capabilities(TENSORFLOW, DEFAULT_TP_MODEL, "v3")
-        mct.get_target_platform_capabilities(TENSORFLOW, TFLITE_TP_MODEL, "v1")
-        mct.get_target_platform_capabilities(TENSORFLOW, QNNPACK_TP_MODEL, "v1")
+        tpc = mct.get_target_platform_capabilities(TENSORFLOW, DEFAULT_TP_MODEL) # Latest
+        self.assertTrue(tpc.version == 'v4')
```
ofirgo, line 239: Maybe use constant for the versions' strings? Not mandatory just think that it would be easier to prevent future errors with all the code duplications we have here.

ofirgo (approved):
Minor, non-mandatory changes requested

tests/keras_tests/test_keras_tp_model.py
reuvenperetz: You're right. It's worth considering enum for that, but let's do it in a different PR.

.github/workflows/run_tests_suite_all_latest_frameworks.yml
reuvenperetz: It was just to test it, I set it to 40min.

PR #394, opened 2022-10-06
Max Cut (Scheduler) algorithm implementation

ofirgo:
Implementing the scheduler algorithm for finding an optimal schedule for running a computation on a model.
The algorithm also provides a mean for calculating the value of a Max Cut in a model computation graph (according to the found schedule).
In addition, the algorithm returns a list of intermediate cuts which are built during the computation, to be used in the future to define and calculate Max Cut KPI for mixed-precision constraint.

The implementation include the following (key) classes and provedures:
- DirectedBipartiteGraph - a base class representation of a bipartite graph.
- MemoryGraph - A bipartite graph which represents the computation options on the actual model graph.
- Cut - An object that represents a cut.
- MemoryElements - a set of activation tensors to be used for defining a cut in the search.
- MaxCutAstar - a class which implements the AStar-like solver for finding a schedule and a max cut.
- compute_graph_max_cut - a wrapper function which runs AStar iterations for finding the optimal schedule.


model_compression_toolkit/core/common/graph/memory_graph/memory_element.py
```diff
@@ -0,0 +1,111 @@
...
+import numpy as np
+
+
+class ActivationMemoryTensor:
+    """
+    A representation of an activation output tensor of a model's layer.
+    """
+
+    def __init__(self, shape: Tuple[Any], node_name: str, node_output_index: int, total_size: float = -1):
+        """
```
lior-dikstein, line 25: Change "total size" to be a boolean flag

model_compression_toolkit/core/common/graph/memory_graph/memory_element.py
```diff
@@ -0,0 +1,111 @@
...
+        Args:
+            shape: The shape of the activation tensor.
+            node_name: The name of the node which its output is represented by the object.
+            node_output_index:
```
lior-dikstein, line 29: Add comments

model_compression_toolkit/core/common/graph/memory_graph/memory_element.py
```diff
@@ -0,0 +1,111 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+from typing import Set, Tuple, Any
+import numpy as np
+
+
+class ActivationMemoryTensor:
+    """
+    A representation of an activation output tensor of a model's layer.
```
lior-dikstein, line 21: Add comment about what is "size"

model_compression_toolkit/core/common/graph/memory_graph/memory_element.py
```diff
@@ -0,0 +1,111 @@
...
+            # MemoryElements are equal if they contain the exact same elements sets
+            return self.elements == other.elements
+        return False
+
+    def __hash__(self):
+        return hash((frozenset(self.elements)))
+
+    def __copy__(self):
+        """
+        Overrides the class copt method.
```
lior-dikstein, line 105: typo in "copy"

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -320,6 +320,19 @@ def get_total_output_params(self):
 
         return sum([np.prod([x for x in output_shape if x is not None]) for output_shape in output_shapes])
 
+    def get_total_input_params(self) -> float:
+        """
+        Calculates the input size of the node.
```
lior-dikstein, line 325: maybe switch the term "size" to num params?

model_compression_toolkit/core/common/graph/memory_graph/compute_graph_max_cut.py
```diff
@@ -0,0 +1,59 @@
...
+    while it < n_iter:
+        estimate = (u_bound + l_bound) / 2
+        schedule, max_cut_size, cuts = max_cut_astar.solve(estimate_factor=estimate, iter_limit=astar_n_iter)
+        if schedule is None:
+            return last_result
+
+        next_u_bound = min(estimate, max_cut_size)
+        last_result = (schedule, max_cut_size, cuts)
+
+        if l_bound >= next_u_bound or l_bound * (1 + eps) > next_u_bound:
```
lior-dikstein, line 54: why do you need the first statement? isn't the second one enough?

model_compression_toolkit/core/common/graph/memory_graph/max_cut_astar.py
```diff
@@ -0,0 +1,414 @@
...
+
+from model_compression_toolkit.core.common import BaseNode
+from model_compression_toolkit.core.common.constants import DUMMY_TENSOR, DUMMY_NODE
+from model_compression_toolkit.core.common.graph.memory_graph.cut import Cut
+from model_compression_toolkit.core.common.graph.memory_graph.memory_element import MemoryElements
+from model_compression_toolkit.core.common.graph.memory_graph.memory_graph import ActivationMemoryTensor, MemoryGraph
+
+
+class DummpyType:
+    """
```
lior-dikstein, line 26: did you mean to name it "Dummpy" or "Dummy"?

model_compression_toolkit/core/common/graph/memory_graph/max_cut_astar.py
```diff
@@ -0,0 +1,414 @@
...
+        while True:
+            yield BaseNode(name=f"{DUMMY_NODE}_{self.counter}",
+                           framework_attr={},
+                           input_shape=tuple(),
+                           output_shape=tuple(),
+                           weights={},
+                           layer_class=DummpyType,
+                           has_activation=False)
+
+            self.counter += 1
```
lior-dikstein, line 52: Maybe the dummy nodes should be placed somewhere else, more common? Because maybe someone will use these in the future in another alg.

model_compression_toolkit/core/common/graph/memory_graph/compute_graph_max_cut.py
ofirgo: You're right, except for the case where eps=0 (I think that in this case the second term would miss the equality that is in the first term).
I'll remove the first term and add equality to the second one to cover all options (it allows a bit broader range but I don't fix it's critical).

model_compression_toolkit/core/common/graph/memory_graph/max_cut_astar.py
ofirgo: I'm not sure that I want to allow someone to use it in another context, this is very specific to this algorithm.

model_compression_toolkit/core/common/graph/memory_graph/memory_graph.py
```diff
@@ -0,0 +1,151 @@
...
+                ot_edges = [oe for oe in out_edges if oe.source_index == i]
+                for oe in ot_edges:
+                    # Add current memory tensor as input to current node's successors
+                    tensor_to_node.append((memory_tensor, oe.sink_node))
+
+        super().__init__(name=model_graph.name + "_memory_graph",
+                         a_nodes=nodes,
+                         b_nodes=memory_tensors,
+                         edges_ab=[nt for nt in node_to_tensor],
+                         edges_ba=[tn for tn in tensor_to_node])
```
lior-dikstein, line 66: why not just pass node_to_tensor and tensor_to_node?

model_compression_toolkit/core/common/graph/memory_graph/memory_graph.py
```diff
@@ -0,0 +1,151 @@
...
+                                 for n in nodes if n in model_graph.get_inputs()]
+
+        nodes_total_memory = [sum([t.total_size for t in self.operation_node_children(n)] +
+                                  [t.total_size for t in self.operation_node_parents(n)])
+                              for n in nodes if n not in model_graph.get_inputs()]
+
+        self.memory_lbound_single_op = max(nodes_total_memory + inputs_tensors_memory)
+
+        self.sources_a = [n for n in self.a_nodes if len(list(self.predecessors(n))) == 0]
+
```
lior-dikstein, line 81: why not use the method "update_sources_a", "update_sinks_b"?

model_compression_toolkit/core/common/graph/memory_graph/memory_graph.py
```diff
@@ -0,0 +1,151 @@
...
+    def update_sinks_b(self):
+        """
+        Updates the list of Side B sinks nodes in the bipartite graph.
+        A node is a sink if it doesn't have any outgoing edges.
+        """
+        self.sinks_b = [n for n in self.b_nodes if len(list(self.successors(n))) == 0]
+
+    def activation_tensor_children(self, activation_tensor: ActivationMemoryTensor) -> List[BaseNode]:
```
lior-dikstein, line 104: the name is a bit confusing. activation tensor is b_node. maybe replace to "b_node_children". same for a_node. because this is the only place you use the terminology "operation node" and "activation_tensor". what do you think?

tests/keras_tests/function_tests/test_graph_max_cut.py
```diff
@@ -0,0 +1,107 @@
+import keras
```
lior-dikstein, line 1: is this only in Keras?

model_compression_toolkit/core/common/graph/memory_graph/memory_graph.py
ofirgo: You're right, I don't know why I did that, fixed.

model_compression_toolkit/core/common/graph/memory_graph/memory_graph.py
ofirgo: I want to keep the terminology of "op node" and "activation tensor" because I think it makes the code more understandable. If it would say "b_node" or "a_node" it is not clear what type of object to expect, and then you need to go to the graph definition and check how "a_nodes" and "b_nodes" are defined.
The terms "operation node" and "activation tensor" appear in the MemoryGraph class description, if you think it's not clear enough I can add more comments about it inside the class but I'm not sure if that would help

tests/keras_tests/function_tests/test_graph_max_cut.py
ofirgo: No, this is a framework-independent code ("common").
I know that usually we test such code under the Keras "function_tests".
The framework is only needed to create a model and feed it to the reader to create a graph.
I can copy the exact same tests to the Pytorch "function_tests" and implement them with a Pytorch model, but I think that this is redundant, since they're not testing framework-dependent functionality.

lior-dikstein (approved):

PR #396, opened 2022-10-19
Disable activation quantization in mixed-precision metric computation for only weight MP search

ofirgo:
Disable activation quantization in mixed-precision metric computation when running mixed precision with target weights KPI only or for nodes that have only single activation candidates (so they are not part of the MP search)

### DIFF
--- a/model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_facade.py
+++ b/model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_facade.py
@@ -84,13 +84,21 @@ def search_bit_width(graph_to_search_cfg: Graph,
         # Since Bit-operations count target KPI is set, we need to reconstruct the graph for the MP search
         graph = substitute(graph, fw_impl.get_substitutions_virtual_weights_activation_coupling())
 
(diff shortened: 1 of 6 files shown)


reuvenperetz (approved):

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_facade.py
```diff
@@ -84,13 +84,21 @@ def search_bit_width(graph_to_search_cfg: Graph,
         # Since Bit-operations count target KPI is set, we need to reconstruct the graph for the MP search
         graph = substitute(graph, fw_impl.get_substitutions_virtual_weights_activation_coupling())
 
+    # If we only run weights compression with MP than no need to consider activation quantization when computing the
+    # MP metric (it adds noise to the computation)
+    disable_activation_for_metric = target_kpi.weights_memory < np.inf and \
```
haihabi, line 89: Why this condition is split into two places one here and the second is https://github.com/sony/model_optimization/pull/396/files#diff-3ab258bd0655a94f106ed17c3ba9090efcfc6772be0223089c4bd22c944f7844 ?

PR #406, opened 2022-11-06
Optimizing distance matrix computation runtime

ofirgo:
Vectorizing the batch loop in the mixed-precision distance matrix computation.
Also, added a few code-level optimizations to the similarity analysis methods to speed up their calculation.

### DIFF
--- a/model_compression_toolkit/core/common/similarity_analyzer.py
+++ b/model_compression_toolkit/core/common/similarity_analyzer.py
@@ -48,7 +48,7 @@ def tensor_norm(x: np.ndarray, p: float = 2.0) -> np.float:
         Lp norm of x.
     """
 
-    return np.power(np.power(np.abs(x), p).sum(axis=-1), 1.0/p)
+    return (np.abs(x) ** p).sum(axis=-1) ** (1.0/p)
 
 
 def flatten_tensor(t: np.ndarray, batch: bool) -> np.ndarray:
@@ -66,7 +66,7 @@ def flatten_tensor(t: np.ndarray, batch: bool) -> np.ndarray:
     if batch:
         f_t = t.reshape([t.shape[0], -1])
     else:
-        f_t = t.reshape([1, -1])
+        f_t = t.flatten()
 
     return f_t
 
--- a/model_compression_toolkit/core/common/similarity_analyzer.py
+++ b/model_compression_toolkit/core/common/similarity_analyzer.py
@@ -37,42 +37,43 @@ def validate_before_compute_similarity(float_tensor: Any, fxp_tensor: Any):
     assert float_tensor.shape == fxp_tensor.shape
 
 
-def tensor_batch_axis(shape: Tuple, batch: bool = False) -> Tuple:
-    """
-    Returns a list of axis for similarity analysis in case we want to run it on a batch
-    instead of a single image/flattened tensor.
-
-    Args:
-        shape: The shape of the data tensor to run similarity analysis on.
-        batch: Whether to run batch similarity analysis or not.
-
-    Returns: If not running on batch, returns None, otherwise, returns a tuple with all axis in the tensor shape except
-     the first one (batch axis).
-
-    """
-
-    return None if not batch else tuple([i for i in range(1, len(shape))])
-
-
-def tensor_norm(x: np.ndarray, p: float = 2.0, batch: bool = False) -> np.float:
+def tensor_norm(x: np.ndarray, p: float = 2.0) -> np.float:
     """
     Compute the Lp-norm of a tensor x.
     Args:
-        x: Tensor to compute its norm
+        x: Flattened tensor to compute its norm
         p: P to use for the Lp norm computation
-        batch: Whether to run batch similarity analysis or not.
 
     Returns:
         Lp norm of x.
     """
 
-    axis = tensor_batch_axis(x.shape, batch)
-    return np.power(np.power(np.abs(x), p).sum(axis=axis), 1.0/p)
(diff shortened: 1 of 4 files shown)


model_compression_toolkit/core/common/similarity_analyzer.py
```diff
@@ -41,74 +41,75 @@ def tensor_norm(x: np.ndarray, p: float = 2.0) -> np.float:
...
-    return np.power(np.power(np.abs(x.flatten()), p).sum(), 1.0/p)
 
+    return np.power(np.power(np.abs(x), p).sum(axis=-1), 1.0/p)
 
-#########################
-# Similarity functions
-#########################
 
-def compute_mse(float_tensor: np.ndarray, fxp_tensor: np.ndarray, norm: bool = False, norm_eps: float = 1e-8) -> float:
+def flatten_tensor(t: np.ndarray, batch: bool) -> np.ndarray:
```
eladc-git, line 54: Why do you need special function and batch flag? why not using np.ravel()?

model_compression_toolkit/core/common/similarity_analyzer.py
```diff
@@ -41,74 +41,75 @@ def tensor_norm(x: np.ndarray, p: float = 2.0) -> np.float:
...
     """
     validate_before_compute_similarity(float_tensor, fxp_tensor)
-    normalized_float_tensor = float_tensor / tensor_norm(float_tensor, 1.0)
-    normalized_fxp_tensor = fxp_tensor / tensor_norm(fxp_tensor, 1.0)
-    return np.mean(np.abs(normalized_float_tensor - normalized_fxp_tensor))
+
+    float_flat = flatten_tensor(float_tensor, batch)
+    fxp_flat = flatten_tensor(fxp_tensor, batch)
+
+    error = ((float_flat - fxp_flat) ** 2).mean(axis=-1)
```
eladc-git, line 101: Maybe here you can reuse code and use tensor_norm.

model_compression_toolkit/core/common/similarity_analyzer.py
```diff
@@ -41,74 +41,75 @@ def tensor_norm(x: np.ndarray, p: float = 2.0) -> np.float:
     """
     Compute the Lp-norm of a tensor x.
     Args:
-        x: Tensor to compute its norm
+        x: Flattened tensor to compute its norm
```
eladc-git, line 44: Norm is useful for all kind of commutation. I think the flatten should be here and not force user to do it outside.

model_compression_toolkit/core/keras/keras_implementation.py
```diff
@@ -451,7 +451,7 @@ def get_node_distance_fn(self, layer_class: type,
             return compute_kl_divergence
         elif layer_class == Dense:
             return compute_cs
-        return lambda x, y: compute_mse(x, y, norm=False, norm_eps=1e-8)
+        return compute_mse
```
eladc-git, line 454: Please add for Pytorch as well.

eladc-git (changes requested):

model_compression_toolkit/core/common/similarity_analyzer.py
ofirgo: This is not a total flattening of the tensor, we flatten here all axis but the first one (batch) in case it is wanted (that is why the batch flag is for).
Correct me if I'm wrong, but np.ravel would return a 1-D array, wouldn't it? is there a way to use it for flattening all axis but the first one?

model_compression_toolkit/core/common/similarity_analyzer.py
ofirgo: I'm not sure that it performs a similar operation - 
(the computation on the i'th row of tensor x)

here: SUM(([x^2]_i)) / |x_i|
tensor_norm: SQRT(SUM([x^2]_i))

Am I missing something?

eladc-git: MSE = norm square. The normalize 1/|x_i| is configurable in mse computation.

model_compression_toolkit/core/common/similarity_analyzer.py
eladc-git: Ok so no ravel because you want to get result per batch. But now tensor_norm return a tensor and not scalar.

ofirgo: If it is preforming the operation per sample than it would return a tensor which is fine (a scalar per sample), otherwise it would return a scalar which is the norm of the entire batch.
I don't understand the problem

model_compression_toolkit/core/common/similarity_analyzer.py
ofirgo: Not sure I understand, what call are you suggesting that I put here? (also, is it necessary? I don't think that the potential code duplication here is problematic, and it might be better to keep the code simple and declarative as it is right now)

eladc-git: Ok. let's leave it.

model_compression_toolkit/core/common/similarity_analyzer.py
eladc-git: If I want to get tensor norm (I don't have a batch dim, only simple tensor), I would expect to get a scalar right, because norm is scalar. In your new tensor_norm function I get a tensor (in previous one I got a scalar). Maybe batch flag which is off by default.

eladc-git (approved):

PR #410, opened 2022-11-09
Add option to data loader to use cached batches (disabled by default)

elad-c:


model_compression_toolkit/core/common/data_loader.py
```diff
@@ -97,10 +110,20 @@ def _sample(self):
...
+            self._sample()
+            data = self.next_batch_data  # get current data
+        else:
+            # get cached batch
+            data = self.cached_batches[self.cache_index]
+            self.cache_index += 1
+            if self.cache_index >= len(self.cached_batches):
+                # when all cached batches were used, initialize index and shuffle the cached batches list
+                self.cache_index = 0
+                random.shuffle(self.cached_batches)
```
haihabi, line 127: Why not shuffle to an image instead of a batch?

PR #411, opened 2022-11-13
Support Pytorch convolution functional API

eladc-git:


model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/const_holder_conv.py
```diff
@@ -0,0 +1,102 @@
...
+        if graph.get_edge_data(first_node,second_node) is None:
+            return graph # skip substitution
+
+        # Set new layer
+        if second_node.type == conv2d:
+            NewLayer = Conv2d
+            out_channel_index = 1
+        elif second_node.type == conv_transpose2d:
+            NewLayer = ConvTranspose2d
+            out_channel_index = 0
```
lior-dikstein, line 63: isn't the out_channel_index framework specific?

eladc-git: I will use fw_info here

PR #464, opened 2022-12-06
Add target platform export package with TF FQ export

reuvenperetz:
Added a new package named target_platform_export. Base Exporter class, BaseKerasExporter, and FakelyQuantKerasExporter were added. Each exporter needs to implement export and save_model methods.
Tests were added as well.


model_compression_toolkit/exporter/target_platform_export/keras/exporters/fakely_quant_keras_exporter.py
```diff
@@ -0,0 +1,124 @@
...
+from model_compression_toolkit.exporter.fully_quantized.keras.builder.quantize_config_to_node import \
+    SUPPORTED_QUANTIZATION_CONFIG
+from model_compression_toolkit.exporter.fully_quantized.keras.quantize_configs.weights_quantize_config import \
+    WeightsQuantizeConfig
+from model_compression_toolkit.exporter.target_platform_export.keras.exporters.base_keras_exporter import BaseKerasExporter
+
+
+class FakelyQuantKerasExporter(BaseKerasExporter):
+    """
+    Exporter for fakely-quant Keras models.
```
ofirgo, line 33: I would suggest adding an extended description in these exporter classes (in general, not just in this one).
The difference between each exporter class is not clear, nor is the purpose of each one of them.
For instance here, the name "fakely-quant" isn't really given any info about the exporter (compared to fully quantized exporter and in general)

model_compression_toolkit/exporter/target_platform_export/keras/exporters/fakely_quant_keras_exporter.py
```diff
@@ -0,0 +1,124 @@
...
+            """
+
+            if isinstance(layer, tfmot.quantization.keras.QuantizeWrapper):
+                # For now, we assume only weights are quantized for models to export. Activation support is in progress.
+                if isinstance(layer.layer, InputLayer):
+                    if not isinstance(layer.quantize_config, NoOpQuantizeConfig):
+                        Logger.error(
+                            f'Currently, FakelyQuantKerasExporter supports weights quantization only. Please consider '
+                            f'disabling activation quantization, until supported ')
+                    return Layer() #Identity
```
ofirgo, line 66: This creates an instance of keras.engine.base_layer, are you sure this equals to an Identity layer? (I don't know I'm just double-chcecking)

model_compression_toolkit/exporter/target_platform_export/keras/exporters/fakely_quant_keras_exporter.py
```diff
@@ -0,0 +1,124 @@
...
+                    with tf.name_scope(new_layer.name):
+                        new_layer.build(layer.input_shape)
+
+                    # Build a list of the layer's new weights.
+                    weights_list = []
+                    for w in new_layer.weights:
+                        val = None
+                        for qw in layer.weights:
+                            if w.name in qw.name:
+                                if w.name.split('/')[-1].split(':')[0] in layer.quantize_config.weight_attrs:
```
ofirgo, line 84: Maybe add a comment for what is the expected patters here so it would be easier to debug in case of failure

model_compression_toolkit/exporter/target_platform_export/keras/exporters/fakely_quant_keras_exporter.py
```diff
@@ -0,0 +1,124 @@
...
+                    return layer.layer
+
+                else:
+                    Logger.error(
+                        f'Currently, FakelyQuantKerasExporter supports weights quantization only. Please consider '
+                        f'disabling activation quantization, until supported ')
+            else:
+                return layer
+
+        # clone each layer in the model and apply _unwrap_quantize_wrappera to layers wrapped with a QuantizeWrapper.
```
ofirgo, line 106: What is "_unwrap_quantize_wrappera"?

tests/keras_tests/function_tests/test_export_keras_fully_quantized_model.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class TestExporter(unittest.TestCase):
+
+    def tearDown(self):
+        os.remove(SAVED_MODEL_PATH)
+
+    def setUp(self) -> None:
+        self.mbv2 = MobileNetV2()
+        self.representative_data_gen = lambda: [np.random.randn(1, 224, 224, 3)]
```
ofirgo, line 41: This seems like a large test, how long does it run?
If it is too long maybe add a unit test (with a small model and input) so it would be simpler to run it in the test suit.

Additionally, you might want to test the exporter with other complicated layers that are not part of MobilenetV2? if so, several small unit tests would do the job I think better than one large MBV2 test.

ofirgo (changes requested):

model_compression_toolkit/exporter/fully_quantized/keras/builder/quantize_config_to_node.py
```diff
@@ -29,6 +29,12 @@
 from model_compression_toolkit.exporter.fully_quantized.keras.quantize_configs.weights_quantize_config import \
     WeightsQuantizeConfig
 
+SUPPORTED_QUANTIZATION_CONFIG = [WeightsQuantizeConfig,
```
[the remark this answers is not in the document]
reuvenperetz: It can not be completely separated since the model exporter expects to receive an exportable model, and part of it is the configurations it uses for wrapping the layers (so these configurations are valid). Since this is the file that generates the configurations in the model_wrapper I think it can fit here

ofirgo (approved):

PR #482, opened 2022-12-13
Modify activation mixed precision tests TPC

ofirgo:
This PR includes the following main changes:
* Updating the tests TPC for activation MP and the tests TPC supporting functions.
* Update function for generating TP model and TPC for MP tests (followed by many imports fixes).
* Fixing failed tests (as a result of the TPC import) and incorrect tests.
* Extended a few tests to improve coverage.


tests/pytorch_tests/tpc_pytorch.py
```diff
@@ -43,44 +59,35 @@ def generate_activation_mp_tpc_pytorch(tp_model, name="activation_mp_pytorch_tp"
                                                     reshape,
                                                     unsqueeze,
                                                     BatchNorm2d,
+                                                    chunk,
```
elad-c, line 62: can we think of a way to use the default TPC from MCT and add necessary changes

elad-c (approved):
Consider refactoring the TPC generation in tests

PR #484, opened 2022-12-14
Mixed precision coverage - final steps

ofirgo:
* Removed unused distance function and added dedicated test for the remaining functions.
* Added error message in bit_width_setter.py and in linear_programming.py for missing condition checks.
* Removed unused code from linear_programming.py.
* Excluded several error log lines from the coverage report using "pragma" comment.
* Extended mixed precision tests for several cases (Keras and Pytorch).
* Additional minor changes for increasing coverage.

### DIFF
--- a/model_compression_toolkit/core/common/mixed_precision/distance_weighting.py
+++ b/model_compression_toolkit/core/common/mixed_precision/distance_weighting.py
@@ -50,34 +50,3 @@ def get_last_layer_weights(distance_matrix: np.ndarray) -> np.ndarray:
     w = np.asarray([0 for _ in range(num_nodes)])
     w[-1] = 1
     return w
-
-
-def get_correlation_to_last_weights(distance_matrix: np.ndarray):
-    """
-    Get weights for weighting the sensitivity among different layers when evaluating MP configurations on
-    model's sensitivity. This function returns weights for each layer, such that the sensitivity
-    is computed using the correlation between each layer and the last layer.
-
-    Args:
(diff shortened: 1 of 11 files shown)


model_compression_toolkit/core/common/mixed_precision/distance_weighting.py
```diff
@@ -50,34 +50,3 @@ def get_last_layer_weights(distance_matrix: np.ndarray) -> np.ndarray:
...
-    num_interest_points = distance_matrix.shape[0]
-    cross_matrix = (np.matmul(ncs_matrix.transpose(), ncs_matrix) / num_samples) / normalization_matrix
-    correlation_to_last_layer = cross_matrix[:, num_interest_points - 1]
-
-    def _softmax(x):
-        e_x = np.exp(x - np.max(x))
-        return e_x / e_x.sum(axis=0)
-
-    return _softmax(correlation_to_last_layer)
-
```
reuvenperetz, line 83: Are we sure we want to remove this?

tests/keras_tests/function_tests/test_lp_search_bitwidth.py
```diff
@@ -111,6 +112,11 @@ def test_search_weights_only(self):
         self.assertTrue(len(bit_cfg) == 1)
         self.assertTrue(bit_cfg[0] == 2)
 
+        target_kpi = None  # target KPI is not defined!
+        with self.assertRaises(Exception):
+            bit_cfg = mp_integer_programming_search(mock_search_manager,
```
reuvenperetz, line 117: Maybe we should add an assertion of the expected error message?

PR #488, opened 2022-12-15
Symmetric lut

elad-c:
support symmetric threshold with LUT quantization

### DIFF
--- a/model_compression_toolkit/core/common/quantization/quantization_fn_selection.py
+++ b/model_compression_toolkit/core/common/quantization/quantization_fn_selection.py
@@ -1,0 +1,50 @@
+# Copyright 2021 Sony Semiconductor Israel, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from collections.abc import Callable
+from functools import partial
+
+from model_compression_toolkit.core.common.target_platform import QuantizationMethod
+from model_compression_toolkit.core.common.quantization.quantizers.kmeans_quantizer import kmeans_quantizer
+from model_compression_toolkit.core.common.quantization.quantizers.lut_kmeans_quantizer import lut_kmeans_quantizer
+from model_compression_toolkit.core.common.quantization.quantizers.uniform_quantizers import power_of_two_quantizer, \
+    symmetric_quantizer, uniform_quantizer
+
+
+def get_weights_quantization_fn(weights_quantization_method: QuantizationMethod) -> Callable:
+    """
+    Generate a function for weight quantization.
+
+    Args:
+        weights_quantization_method: Which quantization method to use for weights.
+    Returns:
+        A function to quantize a weight tensor.
+
+    """
+
+    if weights_quantization_method == QuantizationMethod.POWER_OF_TWO:
+        quantizer_fn = power_of_two_quantizer
+    elif weights_quantization_method == QuantizationMethod.SYMMETRIC:
+        quantizer_fn = symmetric_quantizer
+    elif weights_quantization_method == QuantizationMethod.UNIFORM:
+        quantizer_fn = uniform_quantizer
+    elif weights_quantization_method == QuantizationMethod.KMEANS:
+        quantizer_fn = kmeans_quantizer
+    elif weights_quantization_method in [QuantizationMethod.LUT_POT_QUANTIZER, QuantizationMethod.LUT_SYM_QUANTIZER]:
+        quantizer_fn = lut_kmeans_quantizer
+    else:
+        raise Exception(
+            f'No quantizer function for the configuration of quantization method {weights_quantization_method}')
+    return quantizer_fn
--- a/model_compression_toolkit/core/common/quantization/quantization_fn_selection.py
+++ b/model_compression_toolkit/core/common/quantization/quantization_fn_selection.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Sony Semiconductor Israel, Inc. All rights reserved.
+# Copyright 2022 Sony Semiconductor Israel, Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


model_compression_toolkit/core/common/quantization/quantization_fn_selection.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2021 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo, line 1: year (typo)

model_compression_toolkit/core/common/quantization/quantization_fn_selection.py
```diff
@@ -0,0 +1,50 @@
...
+from functools import partial
+
+from model_compression_toolkit.core.common.target_platform import QuantizationMethod
+from model_compression_toolkit.core.common.quantization.quantizers.kmeans_quantizer import kmeans_quantizer
+from model_compression_toolkit.core.common.quantization.quantizers.lut_kmeans_quantizer import lut_kmeans_quantizer
+from model_compression_toolkit.core.common.quantization.quantizers.uniform_quantizers import power_of_two_quantizer, \
+    symmetric_quantizer, uniform_quantizer
+
+
+def get_weights_quantization_fn(weights_quantization_method: QuantizationMethod) -> Callable:
```
ofirgo, line 26: Consider adding a method for activation quantization function as well (if you already created a separate file for that and moved it for framework info, I think it'll be better to utilize it for activation as well).

model_compression_toolkit/core/common/quantization/quantization_fn_selection.py
```diff
@@ -0,0 +1,50 @@
...
+from collections.abc import Callable
+from functools import partial
+
+from model_compression_toolkit.core.common.target_platform import QuantizationMethod
+from model_compression_toolkit.core.common.quantization.quantizers.kmeans_quantizer import kmeans_quantizer
+from model_compression_toolkit.core.common.quantization.quantizers.lut_kmeans_quantizer import lut_kmeans_quantizer
+from model_compression_toolkit.core.common.quantization.quantizers.uniform_quantizers import power_of_two_quantizer, \
+    symmetric_quantizer, uniform_quantizer
+
+
```
ofirgo, line 25: I don't like the name "selection" here.
It can be confusing with "parameters selection" which uses "QuantizationErrorMethod" enum (too many things that are too similar and it can create confusion).
Consder: "get_quantization_fn" or "quantization_fn_generator" or something like that.

model_compression_toolkit/core/common/quantization/quantization_params_fn_selection.py
```diff
@@ -14,6 +14,7 @@
 # ==============================================================================
```
ofirgo, line 15: I see that here it is also "fn_selection" (so probably that is why you called the new one "selection" as well)
Still think it's confusing since this file actually returns a function for "parameters selection".
So, if you decide to change the name of the other file, maybe change this one too.

model_compression_toolkit/core/common/quantization/quantization_params_generation/symmetric_selection.py
```diff
@@ -52,8 +52,14 @@ def symmetric_selection_tensor(tensor_data: np.ndarray,
         Optimal threshold to quantize the tensor in a symmetric manner.
     """
 
-    unsigned_tensor_data = np.abs(tensor_data)
```
ofirgo, line 55: Why not use the unsigned tensor? (it is debatable what is better for symmetric quantization but it's not necessarily better to take the signed one so I just want to understand the reason)

model_compression_toolkit/core/common/quantization/quantization_params_generation/symmetric_selection.py
```diff
@@ -52,8 +52,14 @@ def symmetric_selection_tensor(tensor_data: np.ndarray,
...
-    tensor_max = get_tensor_max(unsigned_tensor_data, per_channel, channel_axis)
+    if per_channel:
```
ofirgo, line 55: This part is almost duplicated from "get_tensor_max".
Is it possible to update the function itself and continue to use it instead of duplicating most of the code? (or create a "get_symmetric_tensor_max" that uses the "get_tensor_max" function)

model_compression_toolkit/core/common/quantization/quantization_params_generation/symmetric_selection.py
```diff
@@ -52,8 +52,14 @@ def symmetric_selection_tensor(tensor_data: np.ndarray,
...
+        output_shape = get_output_shape(tensor_data.shape, channel_axis)
+        reshaped_tensor_data = reshape_tensor_for_per_channel_search(tensor_data, channel_axis)
+        tensor_max = np.reshape(np.max(reshaped_tensor_data, axis=-1), output_shape)
+        tensor_min = np.reshape(np.min(reshaped_tensor_data, axis=-1), output_shape)
+        tensor_max = np.maximum(-tensor_min, tensor_max * np.power(2.0, n_bits-1) / (np.power(2.0, n_bits-1) - 1))
```
ofirgo, line 60: I don't understand this change. Can we take it offline?

model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
```diff
@@ -62,11 +69,12 @@ def lut_kmeans_tensor(tensor_data: np.ndarray,
     kmeans = KMeans(n_clusters=n_clusters)
     axis_not_channel = [i for i in range(len(tensor_data.shape))]
```
ofirgo, line 70: This variable is no longer being used (can be removed)

model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
```diff
@@ -62,11 +69,12 @@ def lut_kmeans_tensor(tensor_data: np.ndarray,
...
-    if per_channel:
-        thresholds_per_channel = np.max(np.abs(tensor_data), axis=tuple(axis_not_channel), keepdims=True)
-    else:
-        thresholds_per_channel = np.max(np.abs(tensor_data), keepdims=True)
-    thresholds_per_channel = np.power(2.0, np.ceil(np.log2(thresholds_per_channel)))
+
+    threshold_selection_tensor = symmetric_selection_tensor if is_symmetric else power_of_two_selection_tensor
+    thresholds_per_channel = threshold_selection_tensor(tensor_data, p, n_bits, per_channel,
+                                                        channel_axis, n_iter, min_threshold,
+                                                        qc.QuantizationErrorMethod.NOCLIPPING)[THRESHOLD]
```
ofirgo, line 76: Why not use the given quant_error_method instead of no clipping?

ofirgo (changes requested):
Several changes suggestions (some already resolved offline).

model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
elad-c: MSE's may have unexpected bevaiour, so as it is the default, we decided on fixing it to NC for now

ofirgo (approved):

PR #539, opened 2023-01-22
support tensorflow 2.11

eladc-git:


tests/keras_tests/function_tests/test_get_gptq_config.py
```diff
@@ -77,16 +77,18 @@ def test_get_keras_gptq_config(self):
                                                  loss=multiple_tensors_mse_loss,
                                                  rounding_type=RoundingType.GumbelRounding, quantizer_config=gc),
                                ]
-        gptqv2_configurations = [get_keras_gptq_config(n_epochs=1),
-                                 get_keras_gptq_config(n_epochs=1),
-                                 get_keras_gptq_config(n_epochs=1, optimizer=tf.keras.optimizers.Adam())]
+        gptqv2_configurations = [get_keras_gptq_config(n_epochs=1,optimizer=tf.keras.optimizers.Adam()),
+                                 get_keras_gptq_config(n_epochs=1,optimizer=tf.keras.optimizers.Adam()),
+                                 get_keras_gptq_config(n_epochs=1,optimizer=tf.keras.optimizers.Adam())]
+
```
reuvenperetz, line 83: Why is this related to v2.11 specifically?

eladc-git: This is weird test issue. The old version (without optimizer=tf.keras.optimizers.Adam()) uses the same optimizer object in several tests. In 2.11 the optimizer change and they add some extra checks and we fail on some of them. So we need to set object per test and not reuse it.

reuvenperetz (approved):

PR #547, opened 2023-01-25
Add tests for uniform and symmetric activations quantization

lapid92:
Add tests for uniform bounded activation quantization
Add tests for symmetric bounded activation quantization

### DIFF
--- a/model_compression_toolkit/core/common/quantization/quantization_params_generation/qparams_activations_computation.py
+++ b/model_compression_toolkit/core/common/quantization/quantization_params_generation/qparams_activations_computation.py
@@ -66,7 +66,7 @@ def get_activations_qparams(activation_quant_cfg: NodeActivationQuantizationConf
                 activation_quant_cfg.activation_quantization_params_fn = \
                     quantization_params_generation.uniform_no_clipping_selection_min_max
     else:
-        signed = np.any(bins_values[:-1][bins_counts > 0] < 0)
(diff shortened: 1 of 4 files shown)


model_compression_toolkit/core/common/quantization/quantization_params_generation/qparams_activations_computation.py
```diff
@@ -66,7 +66,7 @@ def get_activations_qparams(activation_quant_cfg: NodeActivationQuantizationConf
                 activation_quant_cfg.activation_quantization_params_fn = \
                     quantization_params_generation.uniform_no_clipping_selection_min_max
     else:
-        signed = np.any(bins_values[:-1][bins_counts > 0] < 0)
+        signed = np.any(bins_values[:-1][bins_counts > 0] < 0)  # pragma: no cover
```
haihabi, line 69: Why no pragma?

lapid92: The line execute when prior info hasn't been set.
It's not possible while using the quantization facade, but I'll do a function test.

haihabi: If this case is not feasible then make it an error otherwise test it

ofirgo (approved):

PR #553, opened 2023-01-29
Fix issue in onnx exporter

reuvenperetz:
ONNX opset version 16 is not supported in torch 1.11. This commit adds a fix that sets the correct opset version according to the torch version.


model_compression_toolkit/exporter/model_exporter/pytorch/fakely_quant_onnx_pytorch_exporter.py
```diff
@@ -19,8 +19,13 @@
 from model_compression_toolkit.core.common import Logger
 from model_compression_toolkit.core.pytorch.utils import to_torch_tensor
 from model_compression_toolkit.exporter.model_exporter.pytorch.base_pytorch_exporter import BasePyTorchExporter
+from packaging import version
 
-OPSET_VERSION = 16
+# ONNX opset version 16 is supported from PyTorch 1.12
+if version.parse(torch.__version__) < version.parse("1.12"):
```
eladc-git, line 25: Did you find reference for that? what  about less then op15? Maybe dictionary with versions is more suitable.
I know https://onnxruntime.ai/docs/reference/compatibility.html

eladc-git (approved):

PR #559, opened 2023-02-01
fix include zero QAT keras

eladc-git:


model_compression_toolkit/qat/keras/quantizer/quant_utils.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2021 Sony Semiconductor Israel, Inc. All rights reserved.
```
elad-c, line 1: 2023?

model_compression_toolkit/qat/keras/quantizer/quant_utils.py
```diff
@@ -0,0 +1,48 @@
...
+import tensorflow as tf
+from typing import Tuple
+
+
+def adjust_range_to_include_zero(range_min: tf.Tensor,
+                                 range_max: tf.Tensor,
+                                 n_bits: int) -> Tuple[tf.Tensor, tf.Tensor]:
+    """
+    Adjusting the quantization range to include representation of 0.0 in the quantization grid.
+    If quantization per-channel, then range_min and range_max should be tensors in the specific shape that allows
```
elad-c, line 25: there's no channel axis here, so maybe only write that for per_channel quantization min\max should be a vector

model_compression_toolkit/quantizers_infrastructure/keras/inferable_quantizers/base_uniform_inferable_quantizer.py
```diff
@@ -42,6 +41,9 @@ def __init__(self,
         """
         super(BaseUniformInferableQuantizer, self).__init__(quantization_target=quantization_target)
         self.num_bits = num_bits
+        assert np.all(max_range > min_range), f'Expected max_range to be bigger than min_range!'
+        min_range, max_range = adjust_range_to_include_zero(min_range, max_range, num_bits)
```
elad-c, line 45: Add warning if min_range or max_range changed because of this function.

elad-c (changes requested):

model_compression_toolkit/quantizers_infrastructure/keras/inferable_quantizers/base_uniform_inferable_quantizer.py
eladc-git: Ok but to avoid so many warning I set fix_range in QAT also before calling InferableQuantizer.

elad-c (approved):

PR #561, opened 2023-02-02
fix Keras QAT tests 

eladc-git:
1. Fix DW test
2. Disable loading tests for now (works only in tf2.11)


tests/keras_tests/feature_networks_tests/test_features_runner.py
```diff
@@ -657,15 +657,15 @@ def test_multi_head_attention(self):
                                    separate_key_value=separate_key_value, output_dim=14).run_test()
 
     def test_qat(self):
```
ofirgo, line 659: Please add a TODO indicating that we have a problem with loading in tf versions < 2.10

reuvenperetz (approved):

PR #579, opened 2023-02-13
Integrate GPTQ with Quantization Infra for PyTorch

ofirgo:
Integrating quantization infrastructure into GPTQ (PyTorch) to allow simple use and development of GPTQ quantizers:

* Added BasePytorchGPTQTrainableQuantizer base class for all GPTQ Pytorchquantizers.
* Removed GPTQ model builder and weights config and replaced them with QI's quantization_builder template and trainable quantizer configs.
* Some additional changes including code duplications removal (also, between Keras and Pytorch).


model_compression_toolkit/gptq/pytorch/gptq_training.py
```diff
@@ -83,16 +86,52 @@ def __init__(self,
...
+            node (BaseNode): Node for quantization decision
+        Returns:
+            A boolean whether the layer is to be wrapped with a Quantization Wrapper.
+        """
+
+        if node.is_weights_quantization_enabled() and not self.fw_info.is_kernel_op(node.type):
+            Logger.error("GPTQ Error: Quantizing a node without a kernel isn't supported")
+        return node.is_weights_quantization_enabled()
+
+    def gptq_wrapper(self, n: BaseNode, layer: Module) -> Union[qi.KerasQuantizationWrapper, Module]:
```
elad-c, line 103: keras quantizer in pytorch?

model_compression_toolkit/gptq/pytorch/quantizer/base_pytorch_gptq_quantizer.py
```diff
@@ -0,0 +1,103 @@
...
+
+            Returns: A list with the quantization auxiliary variables.
+
+            """
+
+            return []
+
+        def get_quantization_variable(self) -> List[Tensor]:
+            """
+            This function return a list with the quantizer's quantization parameters variables.
```
elad-c, line 75: returns
parameters or variables?

model_compression_toolkit/gptq/pytorch/quantizer/quantization_builder.py
```diff
@@ -0,0 +1,71 @@
...
+from model_compression_toolkit.quantizers_infrastructure.common.get_quantizer_config import \
+    get_trainable_quantizer_weights_config
+from model_compression_toolkit.qat.pytorch.quantizer.base_pytorch_qat_quantizer import BasePytorchQATTrainableQuantizer
+from model_compression_toolkit.quantizers_infrastructure import QuantizationTarget, BasePyTorchInferableQuantizer
+from model_compression_toolkit.quantizers_infrastructure.common.get_quantizers import get_trainable_quantizer_class, \
+    get_inferable_quantizer_class
+
+
+def quantization_builder(n: common.BaseNode,
+                         gptq_config: GradientPTQConfig,
```
elad-c, line 33: Shouldn't with be GradientPTQConfigV2?

model_compression_toolkit/gptq/pytorch/quantizer/ste_rounding/symmetric_ste.py
```diff
@@ -0,0 +1,200 @@
...
+                                                                  requires_grad=True))
+
+        # save the quantizer added parameters for later calculations
+        self.quantizer_parameters = {PTQ_THRESHOLD: layer.get_parameter(f"{name}_{PTQ_THRESHOLD}"),
+                                     AUXVAR: layer.get_parameter(f"{name}_{AUXVAR}")}
+
+        return self.quantizer_parameters
+
+
+    def get_aux_variable(self) -> List[torch.Tensor]:
```
elad-c, line 136: Ageneral thought: aux variable is very specific to this quantizer. why not call this function get_trainable_variables?

model_compression_toolkit/quantizers_infrastructure/pytorch/quantize_wrapper.py
```diff
@@ -171,6 +171,17 @@ def set_quantize_weights(self, quantized_weights: dict):
                 weight = quantized_weights.get(weight_attr)
                 setattr(self.layer, weight_attr, weight)
 
+        def get_weights_vars(self) -> List[Tuple[str, Any, BaseInferableQuantizer]]:
+            """
+            A getter for the layer's weights variables.
+
+            Returns:
+                List pf tuples of the wrapped layer's weights variables with weight name, values and assigned quantizer.
```
elad-c, line 179: of

elad-c (changes requested):

model_compression_toolkit/gptq/pytorch/quantizer/ste_rounding/symmetric_ste.py
```diff
@@ -0,0 +1,200 @@
...
+from model_compression_toolkit.quantizers_infrastructure.common.base_inferable_quantizer import mark_quantizer
+
+
+def pertubation_symmetric_quantizer(input_tensor: torch.Tensor,
+                                    auxvar_tensor: nn.Parameter,
+                                    max_tensor: torch.Tensor,
+                                    num_bits: int,
+                                    signed: bool,
+                                    power_of_two: bool,
+                                    max_lsbs_change: int = 1) -> nn.Parameter:
```
reuvenperetz, line 40: Set default max_lsbs_change as a constant (here and in quantizer init lambda function)

model_compression_toolkit/gptq/pytorch/quantizer/base_pytorch_gptq_quantizer.py
```diff
@@ -0,0 +1,103 @@
...
+                Keys must match NodeQuantizationConfig attributes
+
+            """
+            weights = {}
+            for weight, quantizer_vars, quantizer in layer.get_weights_vars():
+                weights.update({weight: quantizer(training=False, inputs=quantizer_vars)})
+
+            quant_config = {WEIGHTS_QUANTIZATION_PARAMS: self.get_quant_config()}
+
+            return weights, quant_config, {}
```
reuvenperetz, line 61: Is it possible to remove the empty activation config?

reuvenperetz (changes requested):

model_compression_toolkit/gptq/common/gptq_config.py
```diff
@@ -126,6 +126,22 @@ def _verify_quantizer_config(self, quantizer_config, rounding_type) -> bool:
         return type(quantizer_config) == GPTQQuantizerConfig
 
 
+    def get_extended_quantizer_parametes(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
ofirgo: So, wouldn't it be better to implement it in the parent class, i.e., only in GPTQConfig?
In general, there are already other variables/methods that are only implemented for the new experimental API (not only in GPTQ) and are not supported in the old one.
Maybe it's time to remove the old one..

model_compression_toolkit/gptq/pytorch/graph_info.py
```diff
@@ -12,24 +12,25 @@
...
+from model_compression_toolkit.core.pytorch.default_framework_info import DEFAULT_PYTORCH_INFO
+from model_compression_toolkit.gptq.common.gptq_graph import get_kernel_attribute_name_for_gptq
+from model_compression_toolkit.quantizers_infrastructure import PytorchQuantizationWrapper
 
 
-def get_trainable_parameters(fxp_model: nn.Module,
-                             add_bias: bool = False,
-                             quantization_parameters_learning: bool = False
-                             ) -> (List[nn.Parameter], List[nn.Parameter], List[nn.Parameter]):
+def gptq_get_trainable_parameters(fxp_model: nn.Module,
```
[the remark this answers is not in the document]
ofirgo: I wanted to make it similar to the matching Keras function.
In general, I tried to make the "signature" of matching components similar as possible between frameworks, to prevent confusion.

model_compression_toolkit/gptq/pytorch/quantizer/base_pytorch_gptq_quantizer.py
ofirgo: Currently no. This function is supposed to be implemented in any GPTQ quantizer to allow updating the layers parameters according to the training results.
In theory, we might want to also update activation-related parameters (such as activation threshold that we tried to add before but it didn't work well), so we need to return 3 dictionaries from this method.

model_compression_toolkit/gptq/pytorch/quantizer/base_pytorch_gptq_quantizer.py
ofirgo: These are the variables that represent the quantization parameters.
I would have changed it but this method appears in 5 different locations with this same comment.
So unless you think it is totally confusing, I prefer to not change it to prevent messing up the PR.

model_compression_toolkit/gptq/pytorch/quantizer/base_pytorch_gptq_quantizer.py
```diff
@@ -0,0 +1,103 @@
...
+                A dictionary of attributes the quantize_config retraining has changed during GPTQ retraining.
+                Keys must match NodeQuantizationConfig attributes.
+
+            """
+            raise NotImplemented(f'{self.__class__.__name__} have to implement the '
+                                 f'quantizer\'s get_quant_config.')
+
+else:
+    class BasePytorchGPTQTrainableQuantizer(BasePytorchTrainableQuantizer):
+        def __init__(self,
```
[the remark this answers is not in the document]
ofirgo: Not sure I understand what you mean

model_compression_toolkit/gptq/pytorch/quantizer/quantization_builder.py
ofirgo: Not until we remove the old API (I think..)
Unless we don't want to currently support GPTQ with the old API..

model_compression_toolkit/gptq/pytorch/quantizer/ste_rounding/symmetric_ste.py
```diff
@@ -0,0 +1,200 @@
...
+            quantization_config: Trainable weights quantizer config.
+        """
+        super().__init__(quantization_config)
+        self.num_bits = quantization_config.weights_n_bits
+        self.per_channel = quantization_config.weights_per_channel_threshold
+
+        threshold_values = quantization_config.weights_quantization_params[THRESHOLD]
+        self.threshold_shape = np.asarray(threshold_values).shape
+        self.threshold_values = np.reshape(np.asarray(threshold_values), [-1]) if self.per_channel else float(
+            threshold_values)
```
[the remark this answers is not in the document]
ofirgo: It appears in so many places in the code without an assertion, I think that if this is something that we want to assert we should refactor it code-wide.
In my opinion, this is not needed (if we start to assert things like that we can assert every second line.

model_compression_toolkit/gptq/pytorch/quantizer/ste_rounding/symmetric_ste.py
ofirgo: It seems very specific to this quantizer now, because we don't have many quantizers implemented, but it is a common thing in many quantizers that we might want to implement in the future (it was part of the gumbel quantizer that we removed, and a uniform quantizer that we also removed for now).
I doesn't necessarilu return all trainable parameters so I wouldn't change the name to this.

model_compression_toolkit/quantizers_infrastructure/common/base_trainable_quantizer.py
```diff
@@ -147,3 +147,22 @@ def convert2inferable(self) -> BaseInferableQuantizer:
             BaseInferableQuantizer object.
         """
         raise NotImplemented  # pragma: no cover
+
+    @staticmethod
+    def get_threshold_reshape_shape(tensor_shape, quant_axis, quant_axis_dim):
```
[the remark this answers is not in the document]
ofirgo: Regarding the second quastion - we can put it in utils, but there is not a common quant_util in GPTQ (just per-framework), and also, it is only used inside the quantizers, so I figured putting it in the quantizer's parent class is suitable.

model_compression_toolkit/gptq/pytorch/quantizer/base_pytorch_gptq_quantizer.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import Union, Dict, List
+
+from torch import Tensor
+
+from model_compression_toolkit.core.common.logger import Logger
+from model_compression_toolkit.core.common.constants import FOUND_TORCH
+from model_compression_toolkit.gptq.common.gptq_constants import WEIGHTS_QUANTIZATION_PARAMS
+
+from model_compression_toolkit.quantizers_infrastructure import TrainableQuantizerWeightsConfig, \
+    TrainableQuantizerActivationConfig, BasePytorchTrainableQuantizer
```
reuvenperetz: Move import after if found

reuvenperetz (approved):

elad-c (approved):

PR #589, opened 2023-03-01
Add get_quantized_weights to Keras quantization wrapper

reuvenperetz:


model_compression_toolkit/quantizers_infrastructure/inferable_infrastructure/keras/quantize_wrapper.py
```diff
@@ -314,6 +314,18 @@ def get_weights_vars(self) -> List[Tuple[str, Any, BaseInferableQuantizer]]:
 
             return self._weights_vars
 
+        def get_quantized_weights(self) -> Dict[str, tf.Tensor]:
+            """
+
+            Returns: A dictionary of weights attributes to quantized weights.
+
+            """
```
ofirgo, line 322: you can make it a one-liner:
`return {name: quantizer(w) for name, w, quantizer in self.get_weights_vars()}`

not important but it's a bit cleaner

ofirgo (approved):
What about Pytorch's quantize_wrapper?

PR #591, opened 2023-03-01
Limit ILP runtime

reuvenperetz:
Limit ILP solver to 60 seconds. Use PULP_CBC_CMD as the default solver (as this is PuLP's default solver).

### DIFF
--- a/model_compression_toolkit/core/common/mixed_precision/search_methods/linear_programming.py
+++ b/model_compression_toolkit/core/common/mixed_precision/search_methods/linear_programming.py
@@ -22,6 +22,8 @@ from model_compression_toolkit.core.common import Logger
 from model_compression_toolkit.core.common.mixed_precision.kpi_tools.kpi import KPI, KPITarget
 from model_compression_toolkit.core.common.mixed_precision.mixed_precision_search_manager import MixedPrecisionSearchManager
 
+# Limit ILP solver runtime in seconds
+SOLVER_TIME_LIMIT = 60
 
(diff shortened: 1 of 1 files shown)


model_compression_toolkit/core/common/mixed_precision/search_methods/linear_programming.py
```diff
@@ -64,7 +66,10 @@ def mp_integer_programming_search(search_manager: MixedPrecisionSearchManager,
                                     target_kpi,
                                     search_manager)
 
-    lp_problem.solve()  # Try to solve the problem.
+    # Use default PULP solver. Limit runtime in seconds
+    solver = PULP_CBC_CMD(timeLimit=SOLVER_TIME_LIMIT)
```
ofirgo, line 70: Are you sure that this is the solver that has been running so far? wasn't it running multiple algorithms in order to get the best solution? (I'm just verifying that this wouldn't change the behavior)

ofirgo (approved):

model_compression_toolkit/core/common/mixed_precision/search_methods/linear_programming.py
reuvenperetz: Yeah, I'm sure. PuLP default solver selection process is:

# Default solver selection
if PULP_CBC_CMD().available():
    LpSolverDefault = PULP_CBC_CMD()
elif GLPK_CMD().available():
    LpSolverDefault = GLPK_CMD()
elif COIN_CMD().available():
    LpSolverDefault = COIN_CMD()
else:
    LpSolverDefault = None

PR #595, opened 2023-03-08
Change PyTorch reshape and view operators batch size value

lapid92:
After the fix the batch size would infer from the length of the array and remaining dimensions of the input to the reshape and view operators.


ofirgo (approved):
Discussed fixes offline

PR #630, opened 2023-04-13
fix keras QAT tutorial

eladc-git:


MrLIk:
Hi @eladc-git,
Would you kindly also to add the installation of the tensorflow-model-optimization.
It is necessary because the absence of it causes the following error:
```bash
Exception: Installing tensorflow and tensorflow_model_optimization is mandatory when using keras_quantization_aware_training_init. Could not find Tensorflow package.
```
But it is not sufficient to simply add:
```python
! pip install tensorflow
! pip install tensorflow-model-optimization
! pip install model-compression-toolkit 
```
The simple installation of those libraries as above causes the following error in the second :
```bash
---------------------------------------------------------------------------

RuntimeError                              Traceback (most recent call last)

RuntimeError: module compiled against API version 0x10 but this version of numpy is 0xf

---------------------------------------------------------------------------

ImportError                               Traceback (most recent call last)

[<ipython-input-3-5e76b47e3f81>](https://localhost:8080/#) in <cell line: 4>()
      2 from keras.datasets import mnist
      3 from keras import Model, layers, datasets
----> 4 import model_compression_toolkit as mct
      5 import numpy as np

11 frames

[/usr/local/lib/python3.9/dist-packages/numpy/testing/_private/utils.py](https://localhost:8080/#) in <module>
     21 from numpy.core import(
     22      intp, float32, empty, arange, array_repr, ndarray, isnat, array)
---> 23 import numpy.linalg.lapack_lite
     24 
     25 from io import StringIO

ImportError: numpy.core.multiarray failed to import


---------------------------------------------------------------------------
NOTE: If your import is failing due to a missing package, you can
manually install dependencies using either !pip or !apt.

To view examples of installing some common dependencies, click the
"Open Examples" button below.
---------------------------------------------------------------------------
```
It is caused by the re-installation of numpy library by the tensorflow_model_optimization library.
I fixed it by specifying versions of those libraries:
```bash
! pip install tensorflow==2.7.0
! pip install tensorflow-model-optimization==0.7.0
! pip install model-compression-toolkit 
```
I don't suggest a pull request because I don't sure if this is appropriate variant.

tutorials/example_keras_qat.ipynb
```diff
@@ -36,7 +36,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "! pip install -q tensorflow\n",
+    "! pip install -q tensorflow==2.10.1\n",
+	"! pip install -q tensorflow-model-optimization\n",
```
elad-c, line 40: better set the tf-model-optimization and MCT versions too

elad-c (approved):

PR #633, opened 2023-04-19
Move target platform

elad-c:


model_compression_toolkit/target_platform_capabilities/tpc_models/tflite_tpc/latest/__init__.py
```diff
@@ -13,10 +13,10 @@
 # limitations under the License.
 # ==============================================================================
 from model_compression_toolkit.core.common.constants import FOUND_TF, FOUND_TORCH
-from model_compression_toolkit.tpc_models.qnnpack_tpc.v1.tp_model import get_tp_model, generate_tp_model, get_op_quantization_configs
```
ofirgo, line 16: Same, but on the opposite direction (I'm now realizing that is might just be an issue with how github presents the changes, but I'm keeping the comment so you'll take a look to verify this)

ofirgo (approved):

model_compression_toolkit/target_platform_capabilities/tpc_models/tflite_tpc/latest/__init__.py
elad-c: Looks like it. Seems fine in the code. I'll verify after the merge

PR #634, opened 2023-04-19
Fix cfg creation in weights quantization granularity

reuvenperetz:
When creating quantization configuration for weights the per-tensor/per-channel information was taken from the global quantization configuration instead of TPC. This commit fixes this.


eladc-git (approved):
What about test for pytorch TPC?

reuvenperetz:
@eladc-git  The cfg creation is in common, it framework-agnostic

PR #641, opened 2023-04-27
Add export mode info to tpc

lapid92:


reuvenperetz (commented):
Please make sure the default modes are the actual modes we want to work with

PR #663, opened 2023-05-17
Fix GPTQ issue

reuvenperetz:
Add quantization holder for GPTQ models when activation quantization is needed.

### DIFF
--- a/model_compression_toolkit/core/keras/back2framework/keras_model_builder.py
+++ b/model_compression_toolkit/core/keras/back2framework/keras_model_builder.py
@@ -299,12 +299,10 @@ class KerasModelBuilder(BaseModelBuilder):
                 out_tensors_of_n = op_func(out_tensors_of_n_float)
 
                 # In case the activation quantizer is attached out of the wrapper we use get_activation_quantizer_holder
-                # for the activation quantization holder
-                if self.use_activation_holder_during_model_building:
+                # for the activation quantization holder (if the node's activations are quantized)
+                if n.is_activation_quantization_enabled() and self.use_activation_holder_during_model_building:
                     activation_quantizer_holder = self.get_activation_quantizer_holder(n)
                     out_tensors_of_n = activation_quantizer_holder(out_tensors_of_n)
-
-
 
             elif n.is_activation_quantization_enabled(): # Used only when old exporter is used
                 out_tensors_of_n = self._quantize_node_activations(n, out_tensors_of_n_float)
@@ -333,7 +331,7 @@ class KerasModelBuilder(BaseModelBuilder):
             else:
                 # In case the activation quantizer is attached out of the wrapper we use get_activation_quantizer_holder
                 # for the activation quantization holder
-                if self.use_activation_holder_during_model_building:
+                if n.is_activation_quantization_enabled() and self.use_activation_holder_during_model_building:
                     activation_quantizer_holder = self.get_activation_quantizer_holder(n)
                     out_tensors_of_n = activation_quantizer_holder(out_tensors_of_n_float)
 
--- a/model_compression_toolkit/core/keras/back2framework/keras_model_builder.py
+++ b/model_compression_toolkit/core/keras/back2framework/keras_model_builder.py
@@ -23,6 +23,18 @@ from model_compression_toolkit.core.common.back2framework.base_model_builder imp
 from model_compression_toolkit.core.common.user_info import UserInformation
 from model_compression_toolkit.quantizers_infrastructure.inferable_infrastructure.keras.activation_quantization_holder import ActivationQuantizationHolder
 
+import tensorflow as tf
+from keras.engine.input_layer import InputLayer
+from keras.models import Model, clone_model
+from packaging import version
+
+from model_compression_toolkit.constants import INPUT_BASE_NAME
+from model_compression_toolkit.core.common.back2framework.base_model_builder import BaseModelBuilder
+from model_compression_toolkit.core.common.user_info import UserInformation
+from model_compression_toolkit.quantizers_infrastructure.inferable_infrastructure.keras\
+    .activation_quantization_holder import \
+    ActivationQuantizationHolder
+
 # As from Tensorflow 2.6, keras is a separate package and some classes should be imported differently.
 if version.parse(tf.__version__) < version.parse("2.6"):
     from tensorflow.keras.layers import Input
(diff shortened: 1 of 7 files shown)


model_compression_toolkit/qat/keras/quantization_facade.py
```diff
@@ -60,28 +60,21 @@
...
         """
         if _is_qat_applicable(n, DEFAULT_KERAS_INFO):
             weights_quantizers, activation_quantizers = quantization_builder(n, qat_config, DEFAULT_KERAS_INFO)
-            if include_activation_quantizers:
-                return qi.KerasQuantizationWrapper(layer, weights_quantizers, activation_quantizers)
-            else:
-                return qi.KerasQuantizationWrapper(layer, weights_quantizers)
-        else:
-            return layer
+            return qi.KerasQuantizationWrapper(layer, weights_quantizers)
```
ofirgo, line 76: This change is not related to this PR right?
Is it something that was missing from the PR that added the holder to QAT?

model_compression_toolkit/qat/keras/quantizer/quantization_builder.py
```diff
@@ -40,20 +40,22 @@ def get_activation_quantizer_holder(n: common.BaseNode,
...
-        # Holder by definition uses a single quantizer for the activation quantization
-        # thus we make sure this is the only possible case (unless it's a node with no activation
-        # quantization, which in this case has an empty list).
-        if len(activation_quantizers) > 0:
-            assert len(activation_quantizers) == 1, f'ActivationQuantizationHolder supports a ' \
-                                                    f'single quantizer but {len(activation_quantizers)} quantizers ' \
-                                                    f'were found for node {n}'
-            return ActivationQuantizationHolder(activation_quantizers[0])
-    return None
+    # if _is_qat_applicable(n, DEFAULT_KERAS_INFO):
```
ofirgo, line 43: Remove comment.

model_compression_toolkit/qat/keras/quantizer/quantization_builder.py
```diff
@@ -40,20 +40,22 @@ def get_activation_quantizer_holder(n: common.BaseNode,
...
+    _, activation_quantizers = quantization_builder(n,
```
ofirgo, line 44: Also, is this another remainder from the previous "adding holder to QAT" PR?

model_compression_toolkit/gptq/keras/graph_info.py
```diff
@@ -50,6 +50,7 @@ def get_gptq_trainable_parameters(fxp_model: Model,
                                                                   fw_info=DEFAULT_KERAS_INFO)
 
             # collect trainable weights per quantizer
+            assert kernel_attribute in layer.weights_quantizers, f'{kernel_attribute} was not found in weight quantizers of layer {layer.layer}'
```
ofirgo, line 53: Why not logger error?

model_compression_toolkit/gptq/keras/gptq_training.py
```diff
@@ -133,28 +133,21 @@ def _is_gptq_applicable(self,
...
-            include_activation_quantizers: Whether to use the wrapper for the activation quantizer or not
             n: A node of mct graph.
             layer: A keras layer
 
         Returns: Wrapped layer if the layer should be wrap, otherwise returns the layer as is.
 
         """
-        if self._is_gptq_applicable(n):
+        if self._is_gptq_weights_trainable(n):
             weights_quantizers, activation_quantizers = quantization_builder(n, self.gptq_config)
```
ofirgo, line 148: maybe replace activation_quantizers  with '_' since this variable is not being used (and it can be confusing)

model_compression_toolkit/gptq/keras/gptq_training.py
```diff
@@ -169,16 +162,17 @@ def get_activation_quantizer_holder(self, n: common.BaseNode) -> Union[None, Cal
...
-            _, activation_quantizers = quantization_builder(n, self.gptq_config)
-            # Holder by defenition uses a single quantizer for the activation quantization
-            # thus we make sure this is the only possible case (unless it's a node we no activation
-            # quantization, which in this case has an empty list).
-            if len(activation_quantizers) > 0:
-                assert len(activation_quantizers) == 1, f'ActivationQuantizationHolder supports a ' \
-                                                        f'single quantizer but {len(activation_quantizers)} quantizers were found for node {n}'
-                return ActivationQuantizationHolder(activation_quantizers[0])
-        return None
+        _, activation_quantizers = quantization_builder(n, self.gptq_config)
```
ofirgo, line 165: Just a suggestion (need to think about that) - we actually call quantization_builder twice for each node (here and in gptq_wrapper) but each time only take one out of the weights/activation quantizers.
Can we reduce it to a single call?
Maybe call it in the __init__ method of the class and save a mapping between the node to each type of quantizers?

model_compression_toolkit/gptq/keras/gptq_training.py
```diff
@@ -169,16 +162,17 @@ def get_activation_quantizer_holder(self, n: common.BaseNode) -> Union[None, Cal
...
+        # Holder by definition uses a single quantizer for the activation quantization
+        # thus we make sure this is the only possible case (unless it's a node we no activation
+        # quantization, which in this case has an empty list).
+        if len(activation_quantizers) > 0:
+            assert len(activation_quantizers) == 1, f'ActivationQuantizationHolder supports a ' \
```
ofirgo, line 170: Why not logger error?
Also, I would change it to an if-else statement:
`if len == 0
elif len == 1
else (raise error)`

this is clearer in my opinion

model_compression_toolkit/gptq/keras/gptq_training.py
```diff
@@ -169,16 +162,17 @@ def get_activation_quantizer_holder(self, n: common.BaseNode) -> Union[None, Cal
...
+                                                    f'single quantizer but {len(activation_quantizers)} quantizers were found for node {n}'
+            return ActivationQuantizationHolder(activation_quantizers[0])
+
+        # Return identity function if no quantization is needed
+        return lambda x: x
```
ofirgo, line 175: Let's go over this one offline again, I think it's a bad idea to have such a hidden "solution" in our code, to a problem that was created by our design.

model_compression_toolkit/core/keras/back2framework/keras_model_builder.py
```diff
@@ -303,10 +302,9 @@ def _run_operation(self,
                 # for the activation quantization holder
                 if self.use_activation_holder_during_model_building:
                     activation_quantizer_holder = self.get_activation_quantizer_holder(n)
+                    out_tensors_of_n = activation_quantizer_holder(out_tensors_of_n)
```
ofirgo, line 305: I think something is missing, what if we don't have a wrapper but we want to use the holder for activations? why do they need to be coupled like this? (I had something else in mind when we talked about this code yesterday).

Also, if the code for input is similar to the code for the other types of nodes, please extract it to a function, so we won't need to modified twice each time we change something here.

tests/keras_tests/function_tests/test_activation_quantization_holder_gptq.py
```diff
@@ -43,26 +51,32 @@ class TestGPTQModelBuilderWithActivationHolder(unittest.TestCase):
     def test_adding_holder_instead_quantize_wrapper(self):
         input_shape = (8, 8, 3)
         gptq_model = self._get_gptq_model(input_shape, basic_model)
-        self.assertTrue(isinstance(gptq_model.layers[3], ActivationQuantizationHolder))
-        self.assertTrue(isinstance(gptq_model.layers[3].activation_holder_quantizer, ActivationPOTInferableQuantizer))
+        self.assertTrue(isinstance(gptq_model.layers[4], ActivationQuantizationHolder))
```
ofirgo, line 54: How did we end up with only this test that needs index changing?
Let's go over it offline

ofirgo (changes requested):

model_compression_toolkit/qat/keras/quantizer/quantization_builder.py
reuvenperetz: Yes. Removing the flag in the model builder breaks the QAT so it has to be removed here as well.

ofirgo (approved):

PR #680, opened 2023-06-01
QI refactor for PTQ

reuvenperetz:
1. Use a quantizer holder for activations instead of a wrapper during PTQ.
2. In the Keras model builder - replace layers wrapping using clone_model with building wrapped layers in OperationHandler.
3. Remove the Identity layer added for models to overcome the issue of quantizing inputs (now the holder quantizes it, and no wrapper is needed).

Exporter changes:
1. Update substitutions in Keras exporter (to stop handling activation quantizers in the wrapper and take input/output shapes from the wrapped layer since they are now inferred and not the internal layers).
2. Update the Keras validation function to consider new restrictions.

### DIFF
--- a/model_compression_toolkit/exporter/model_exporter/keras/fakely_quant_keras_exporter.py
+++ b/model_compression_toolkit/exporter/model_exporter/keras/fakely_quant_keras_exporter.py
@@ -69,7 +69,9 @@ class FakelyQuantKerasExporter(BaseKerasExporter):
                 Layer after unwrapping.
 
             """
-            assert self.is_layer_exportable_fn(layer), f'Layer {layer.name} is not exportable.'
+
+            # Assert each layer is exportable
+            self.is_layer_exportable_fn(layer)
 
             # If weights are quantized, use the quantized weight for the new built layer.
             if isinstance(layer, KerasQuantizationWrapper):
--- a/model_compression_toolkit/exporter/model_exporter/keras/fakely_quant_keras_exporter.py
+++ b/model_compression_toolkit/exporter/model_exporter/keras/fakely_quant_keras_exporter.py
@@ -72,48 +72,42 @@ class FakelyQuantKerasExporter(BaseKerasExporter):
             assert self.is_layer_exportable_fn(layer), f'Layer {layer.name} is not exportable.'
 
             # If weights are quantized, use the quantized weight for the new built layer.
-            if layer.is_weights_quantization:
-                new_layer = layer.layer.__class__.from_config(layer.layer.get_config())
-                with tf.name_scope(new_layer.name):
-                    new_layer.build(layer.input_shape)
+            if isinstance(layer, KerasQuantizationWrapper):
+                if layer.is_weights_quantization:
+                    new_layer = layer.layer.__class__.from_config(layer.layer.get_config())
+                    with tf.name_scope(new_layer.name):
+                        new_layer.build(layer.input_shape)
 
-                # Build a list of the layer's new weights.
-                weights_list = []
-                # Go over weights, check if they should be quantized, and quantize if this is the case:
-                for w in new_layer.weights:
-                    val = None
-                    for qw in layer.weights:
-                        if w.name in qw.name:
-                            # Use quantized weight if layer attribute should be quantized.
-                            # For example: check if 'kernel_0' is an attribute
-                            # that should be quantized. First, extract 'kernel' from variable name, check if the
-                            # quantize config contains this as an attribute for quantization. If so -
-                            # Take the quantized weight from the quantize_config and set it to the new layer.
-                            attribute_name = w.name.split('/')[-1].split(':')[0]
-                            if attribute_name in layer.weights_quantizers.keys():
-                                quantizer = layer.weights_quantizers.get(attribute_name)
-                                val = quantizer(qw)
-                            else:
-                                val = qw
-                    if val is None:
-                        Logger.error(f'Could not match weight name: {w.name}')
-                    weights_list.append(val)
+                    # Build a list of the layer's new weights.
+                    weights_list = []
+                    # Go over weights, check if they should be quantized, and quantize if this is the case:
+                    for w in new_layer.weights:
+                        val = None
+                        for qw in layer.weights:
+                            if w.name in qw.name:
+                                # Use quantized weight if layer attribute should be quantized.
+                                # For example: check if 'kernel_0' is an attribute
+                                # that should be quantized. First, extract 'kernel' from variable name, check if the
+                                # quantize config contains this as an attribute for quantization. If so -
+                                # Take the quantized weight from the quantize_config and set it to the new layer.
+                                attribute_name = w.name.split('/')[-1].split(':')[0]
+                                if attribute_name in layer.weights_quantizers.keys():
+                                    quantizer = layer.weights_quantizers.get(attribute_name)
+                                    val = quantizer(qw)
+                                else:
+                                    val = qw
+                        if val is None:
+                            Logger.error(f'Could not match weight name: {w.name}')
+                        weights_list.append(val)
 
-                new_layer.set_weights(weights_list)
-                new_layer.trainable = False
+                    new_layer.set_weights(weights_list)
+                    new_layer.trainable = False
 
-                # If activations are also quantized, wrap the layer back using ActivationQuantizeConfig
-                # from original wrapper (weights wrapping is no longer needed).
-                if layer.is_activation_quantization:
-                    new_layer = KerasQuantizationWrapper(layer=new_layer,
-                                                         activation_quantizers=layer.activation_quantizers)
-
-                return new_layer
+                    return new_layer
 
             # If this is a layer with activation quantization only, just return it
             # as activation quantization in the fake-quant case uses the wrapper for quantization.
             return layer
-
 
         # clone each layer in the model and apply _unwrap_quantize_wrapper to layers wrapped with a QuantizeWrapper.
         self.exported_model = tf.keras.models.clone_model(self.model,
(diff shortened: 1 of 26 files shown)


model_compression_toolkit/exporter/model_exporter/keras/int8_tflite_exporter.py
```diff
@@ -93,7 +93,7 @@ def _get_pointwise_layer_to_replace_dense(self, wrapped_layer: KerasQuantization
 
         # Create the point-wise layer
         pw_layer = Conv2D(**pw_cfg)
-        pw_layer.build(wrapped_layer.layer.input_shape)
+        pw_layer.build(wrapped_layer.input_shape)
```
ofirgo, line 96: was this a bug or is something in the wrapper changed that I'm missing? (is there a new field 'input_shape' of the wrapper instead of the layer?)

model_compression_toolkit/exporter/model_exporter/keras/int8_tflite_exporter.py
```diff
@@ -140,31 +140,32 @@ def _get_pointwise_layer_to_replace_dense(self, wrapped_layer: KerasQuantization
...
+            Reshape(wrapped_layer.output_shape[1:])
         ])
 
     def export(self) -> None:
         """
         Export a fully quantized model to its int8 tflite model.
         """
 
-        def _substitute_model(wrapped_layer: KerasQuantizationWrapper) -> keras.layers.Layer:
+        def _substitute_model(layer_to_substitue: keras.layers.Layer) -> keras.layers.Layer:
```
ofirgo, line 157: I'm probably missing something, I don't understand why the wrapper has been replaced with a layer.
Also, isn't it too wide condition? Isn't most Keras layers are of type keras.layers.Layer?

model_compression_toolkit/exporter/model_wrapper/keras/validate_layer.py
```diff
@@ -39,40 +38,34 @@ def is_keras_layer_exportable(layer: Any) -> bool:
         if isinstance(layer, InputLayer):
             return True
 
-        valid_layer = isinstance(layer, KerasQuantizationWrapper)
+        valid_layer = isinstance(layer, Layer)
```
ofirgo, line 41: Same. I don't understand this change.

tests/keras_tests/feature_networks_tests/feature_networks/mixed_precision_tests.py
```diff
@@ -110,13 +111,20 @@ def verify_quantization(self, quantized_model, input_x, weights_layers_idx, weig
...
         get_outputs = K.function([inp], out)
         layer_outs = get_outputs([input_x])
+
         # verifying fake quant nodes output
-        for idx in activation_layers_idx:
-            self.unit_test.assertTrue(np.unique(layer_outs[idx].flatten()).shape[0] <= unique_tensor_values)
+        for layer_out in layer_outs:
+            self.unit_test.assertTrue(np.unique(layer_out).flatten().shape[0] <= unique_tensor_values)
+
+        # model_output = quantized_model(input_x)
```
ofirgo, line 124: Remove?

tests/keras_tests/feature_networks_tests/feature_networks/mixed_precision_tests.py
```diff
@@ -386,13 +402,15 @@ def create_networks(self):
         inputs = layers.Input(shape=self.get_input_shapes()[0][1:])
         x = layers.Conv2D(32, 4)(inputs)
         x = layers.Add()([x, x])
+        # x = layers.Layer()(inputs)
```
ofirgo, line 405: Remove

tests/keras_tests/feature_networks_tests/feature_networks/mixed_precision_tests.py
```diff
@@ -110,13 +111,20 @@ def verify_quantization(self, quantized_model, input_x, weights_layers_idx, weig
                         0] <= unique_tensor_values)
 
         # verify activation quantization
+        holder_layers = get_layers_from_model_by_type(quantized_model, KerasActivationQuantizationHolder)[1:] # skip the input layer
```
ofirgo, line 114: Why do we skip the input layer? (I believe it's related to the error you were getting with the disconnected graph? can you explain how you managed to solve it?)

tests/keras_tests/feature_networks_tests/feature_networks/nested_networks/nested_test.py
```diff
@@ -87,9 +87,9 @@ def compare(self, quantized_model, float_model, input_x=None, quantization_info=
             else:
                 self.unit_test.assertFalse(isinstance(l, Functional) or isinstance(l, Sequential))
         if self.is_inner_functional:
-            num_layers = 9
+            num_layers = 8
```
ofirgo, line 90: Why do we need to change this?
Is it something that is not stable in the test like the referencing to layers by index? If so, we might want to change the test so we won't need to update this everytime there is a change in the graph.

tests/keras_tests/feature_networks_tests/feature_networks/qat/qat_test.py
```diff
@@ -125,7 +125,7 @@ def compare(self, quantized_model, float_model, loaded_model, input_x=None, quan
             dw_weight = float_model.layers[1].weights[0].numpy()
             quantized_dw_weight = dw_layers[0].weights[0].numpy()
         else:
-            self.unit_test.assertTrue(isinstance(quantized_model.layers[2], KerasActivationQuantizationHolder))
+            self.unit_test.assertTrue(isinstance(quantized_model.layers[1], KerasActivationQuantizationHolder))
```
ofirgo, line 128: Referencing by index... is there no workaround for this test?

ofirgo (changes requested):

model_compression_toolkit/exporter/model_wrapper/keras/builder/fully_quantized_model_builder.py
```diff
@@ -32,15 +33,41 @@ def _get_wrapper(node: common.BaseNode,
...
         Returns: Wrapped layer with weights quantizers and activation quantizers
 
         """
-        weights_quantizers, activation_quantizers = get_quantization_quantizers(node)
-        return KerasQuantizationWrapper(layer, weights_quantizers, activation_quantizers)
+        weights_quantizers, _ = get_quantization_quantizers(node)
+        if len(weights_quantizers) > 0:
+            return KerasQuantizationWrapper(layer,
+                                            weights_quantizers)
+        return layer
```
eladc-git, line 46: output type missing a simple layer

model_compression_toolkit/exporter/model_exporter/keras/fakely_quant_keras_exporter.py
```diff
@@ -72,49 +72,43 @@ def _unwrap_quantize_wrapper(layer: Layer):
             assert self.is_layer_exportable_fn(layer), f'Layer {layer.name} is not exportable.'
```
eladc-git, line 72: this assert is now meaningless because you have checks in is_layer_exportable_fn

tests/keras_tests/feature_networks_tests/feature_networks/shift_neg_activation_test.py
```diff
@@ -126,14 +127,14 @@ def compare(self, quantized_model, float_model, input_x=None, quantization_info=
         self.unit_test.assertTrue(float_model.output.shape.as_list() == quantized_model.output.shape.as_list(),
                                   msg=f'Outputs shape mismatch: {float_model.output.shape} != {quantized_model.output.shape}')
 
-        non_linear_layer_fake_quant = quantized_model.layers[2]
-        non_linear_nbits = non_linear_layer_fake_quant.activation_quantizers[0].get_config()['num_bits']
+        non_linear_layer_fake_quant = quantized_model.layers[3]
```
eladc-git, line 130: Did you solve the index issue or just leave it?

eladc-git (changes requested):

model_compression_toolkit/exporter/model_exporter/keras/int8_tflite_exporter.py
reuvenperetz: Now, the model builder uses the wrapped layers for building the model (instead of internal layers). The wrapped layers are used for inference thus they have the input_shape (instead of the internal layers).

model_compression_toolkit/exporter/model_exporter/keras/int8_tflite_exporter.py
reuvenperetz: Before this change, the exporter expected all layers (except for input layers) to be wrapped. After this change, this is not the case as layers that are not quantized, for example, are not wrapped. 
This is not a condition. This is the expected type of input, and all inputs *should* be Keras layers.

model_compression_toolkit/exporter/model_wrapper/keras/validate_layer.py
reuvenperetz: As part of the exporter's validation of a model, it should ensure all layers are of this type. For example, it does not support layers that are sequential models. For this reason, it validates it.

tests/keras_tests/feature_networks_tests/feature_networks/mixed_precision_tests.py
reuvenperetz: It was solved by building the model using the wrapped layers (instead of building and wrapping it separately). Even before this change, the input layers were not checked (see the activation indices that were used).

tests/keras_tests/feature_networks_tests/feature_networks/nested_networks/nested_test.py
reuvenperetz: This is because we're not using identity layers as before. So one layer was removed.

ofirgo (approved):

model_compression_toolkit/exporter/model_wrapper/keras/builder/fully_quantized_model_builder.py
eladc-git: what about this?

eladc-git (approved):

PR #684, opened 2023-06-04
Fix tf export of reused layers

reuvenperetz:
In TF export, as part of weights building of a layer, the attribute input_shape of the layer was searched. When a layer has multiple inbound layers with different shapes this is impossible. For this reason, we get the first input shape of the layer to support cases where multiple input shapes exist.

### DIFF
--- a/tests/keras_tests/exporter_tests/tflite_fake_quant/networks/conv2d_test.py
+++ b/tests/keras_tests/exporter_tests/tflite_fake_quant/networks/conv2d_test.py
@@ -62,3 +62,30 @@ class TestConv2DTFLiteFQExporter(TFLiteFakeQuantExporterBaseTest):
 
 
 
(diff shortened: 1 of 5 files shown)


tests/keras_tests/exporter_tests/tflite_fake_quant/networks/conv2d_test.py
```diff
@@ -62,3 +62,30 @@ def run_checks(self):
 
 
 
+class TestConv2DReusedTFLiteFQExporter(TFLiteFakeQuantExporterBaseTest):
```
elad-c, line 65: consider comparing the outputs of the float and exported networks

elad-c (changes requested):

elad-c (approved):

PR #686, opened 2023-06-06
Support export of QAT finalized models

reuvenperetz:
Two changes to support exporting QAT finalized models:
1. Creating new KerasActivationQuantizationHolder layers after converting the trainable quantizer to inferable. This is done to remove the training-related weights.
2. When creating an unwrap layer (during export), we create the weights based on the wrapped layer instead of going over the weights of the wrapped layer (which caused issues when the model was in eager mode).

### DIFF
--- a/model_compression_toolkit/qat/keras/quantization_facade.py
+++ b/model_compression_toolkit/qat/keras/quantization_facade.py
@@ -260,10 +260,11 @@ if FOUND_TF:
                 layer.convert_to_inferable_quantizers()
             # In the KerasActivationQuantizationHolder case - converting the quantizers only
             # is not enough. We need to create a new layer with inferable quantizers. The reason for that
-            # is that if we only convert the quantizers, the layer will have some weights that do not
-            # match the configuration, thus loading such a model will fail. To overcome this, the
-            # convert_to_inferable_quantizers of KerasActivationQuantizationHolder creates a new layer
-            # from its new configuration after converting the trainable quantizer to an inferable quantizer.
+            # is that if we only convert the quantizers, the layer will have some weights (such as min, max,
+            # threshold) that do not match the configuration, thus loading such a model will fail.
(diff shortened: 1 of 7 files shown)


model_compression_toolkit/qat/keras/quantization_facade.py
```diff
@@ -256,8 +256,16 @@ def keras_quantization_aware_training_finalize(in_model: Model) -> Model:
 
          """
         def _export(layer):
-            if isinstance(layer, (KerasQuantizationWrapper, KerasActivationQuantizationHolder)):
+            if isinstance(layer, KerasQuantizationWrapper):
                 layer.convert_to_inferable_quantizers()
+            # In the KerasActivationQuantizationHolder case - converting the quantizers only
+            # is not enough. We need to create a new layer with inferable quantizers. The reason for that
+            # is that if we only convert the quantizers, the layer will have some weights that do not
+            # match the configuration, thus loading such a model will fail. To overcome this, the
```
ofirgo, line 264: Specify what are these "some weights" that the layer would have (that they are for training only)

tests/keras_tests/feature_networks_tests/feature_networks/network_editor/edit_error_method_test.py
```diff
@@ -62,6 +62,7 @@ def compare(self, quantized_model, float_model, input_x=None, quantization_info=
         holder_layer = get_layers_from_model_by_type(quantized_model, KerasActivationQuantizationHolder)[0]
         input_q_params = holder_layer.activation_holder_quantizer.get_config()
         threshold = input_q_params['threshold']
-        self.unit_test.assertTrue(threshold == 2,
+        self.unit_test.assertTrue(len(threshold) == 1, f'In per-tensor quantization, expected to fins a single threshold but founod {len(threshold)}')
```
ofirgo, line 65: typos ("fins" and "founod")

ofirgo (approved):

PR #689, opened 2023-06-11
Prevent scale of symmetric soft quantizer to be non-positive, when using quantization parameter learning

ofirgo:


model_compression_toolkit/core/common/quantization/quantizers/uniform_quantizers.py
```diff
@@ -24,9 +24,10 @@
 def threshold_is_power_of_two(threshold: np.ndarray, per_channel: bool) -> bool:
     if per_channel:
         thresholds_per_channel = threshold.flatten()
-        return (np.log2(thresholds_per_channel) == list(map(int, np.log2(thresholds_per_channel)))).all()
+        return (thresholds_per_channel > 0).all() \
+               and (np.log2(thresholds_per_channel) == list(map(int, np.log2(thresholds_per_channel)))).all()
 
-    return np.log2(threshold) == int(np.log2(threshold))
+    return threshold > 0 and np.log2(threshold) == int(np.log2(threshold))
```
reuvenperetz, line 30: I think it may be better to do this check separately (otherwise, the function name is wrong)

reuvenperetz (approved):

PR #714, opened 2023-06-20
add informative log for keras de-serialization errors

eladc-git:


model_compression_toolkit/core/keras/back2framework/instance_builder.py
```diff
@@ -84,7 +85,12 @@ def node_builder(n: common.BaseNode) -> Layer:
     framework_attr = copy.copy(n.framework_attr)
     _layer_class = n.layer_class
     framework_attr[LAYER_NAME] = n.name  # Overwrite framework name to identical graph node name
-    node_instance = _layer_class.from_config(framework_attr)  # Build layer from node's configuration.
+    try:
+        node_instance = _layer_class.from_config(framework_attr)  # Build layer from node's configuration.
+    except:
+        Logger.error(
+            f"Keras can not de-serialize layer {_layer_class} attribute {framework_attr} in order to build static graph representation. This is probably because "
```
reuvenperetz, line 92: I would remove the "This is probably." It may happen for many reasons.
Also, attach the exception error to the error message.

reuvenperetz: Also, take care of the coverage

elad-c: I think `framework_attr` might be a large dictionary, so maybe it's better to leave it out of the message.

eladc-git: We do want to mention the "custom layers" to inform the user this is the reason we crash, and not because of something we should have supported

elad-c (approved):

PR #715, opened 2023-06-21
Change get_inferable_quantizers_kwargs argument to get node configuration instead of node

ofirgo:
Change get_inferable_quantizers_kwargs argument to get node configuration instead of node, so it can be broadly used outside the exporter


model_compression_toolkit/exporter/model_wrapper/keras/builder/node_to_quantizer.py
```diff
@@ -24,54 +26,59 @@
...
-                    qi_keras_consts.INPUT_RANK: len(node_w_qc.weights_quantization_params[SCALE_PER_CHANNEL].shape)}
+                    qi_keras_consts.INPUT_RANK: len(node_qc.weights_quantization_params[SCALE_PER_CHANNEL].shape)}
 
         else:
             Logger.critical(f'Not supported quantization method for inferable quantizers.')  # pragma: no cover
 
     elif quantization_target == QuantizationTarget.Activation:
-        # Get the activation quantization configuration for the node
-        node_qc = node.final_activation_quantization_cfg
+        if not isinstance(node_qc, NodeActivationQuantizationConfig):
```
elad-c, line 79: 1. No need for the same check in the weight quantizer section?
2. add no-cover pragma

elad-c (changes requested):

model_compression_toolkit/exporter/model_wrapper/keras/builder/node_to_quantizer.py
ofirgo: 1. There is (line 44)
2. V

elad-c (approved):

PR #719, opened 2023-06-25
Add final filtering phase to node candidates filtering

ofirgo:
Adding an additional filtering step to node candidates filtering, since after first filter we might get duplications in the remaining candidates (in specific mixed precision scenario that include LUT quantization option).

In addition, a small fix to LUT kmeans call is done to eliminate some sklearn's warning

### DIFF
--- a/model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
+++ b/model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
@@ -15,6 +15,8 @@
 import copy
 from typing import List
 
+from model_compression_toolkit.target_platform_capabilities.target_platform import QuantizationMethod
+
 from model_compression_toolkit.core.common import Graph, BaseNode
 from model_compression_toolkit.constants import FLOAT_BITWIDTH
 from model_compression_toolkit.core.common.quantization.candidate_node_quantization_config import \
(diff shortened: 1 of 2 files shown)


model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
```diff
@@ -81,5 +114,8 @@ def filter_node_candidates(node: BaseNode) -> List[CandidateNodeQuantizationConf
 
         for c in filtered_candidates:
             c.weights_quantization_cfg.weights_n_bits = FLOAT_BITWIDTH
+            c.weights_quantization_cfg.weights_quantization_method = QuantizationMethod.POWER_OF_TWO
+
+        final_candidates = _filter_bit_method_dups(filtered_candidates)
```
eladc-git, line 119: why you are using **_filter_bit_method_dups** in 2 places instead one instance before the final return? I guess it's not suppose to effect the single_dummy_candidate.

ofirgo: there is another case that neither of the "if" statements is executed, then an outer call to _filter_bit_method_dups would run on the entire candidates list.
It's not supposed to affect the outcome but I wouldn't want to run something that is not necessary

eladc-git (approved):

PR #748, opened 2023-07-19
Raise error when a Keras model has a custom layer

reuvenperetz:


model_compression_toolkit/core/keras/reader/node_builder.py
```diff
@@ -17,6 +17,9 @@
 import tensorflow as tf
 from packaging import version
 
+from mct_quantizers.logger import Logger
```
elad-c, line 20: Why get the Logger from `mct_quantizers`, and not from `model_compression_toolkit.logger`

elad-c (approved):

PR #763, opened 2023-08-09
Update TPCs with no quantization of operator tf.unstack

lapid92:


elad-c (commented):
Why not add the `tf.unstack` equivalent in pytorch (`torch.unbind`) to all pytorch TPCs?
Also consider adding other operations from [Indexing, Slicing, Joining, Mutating Ops](https://pytorch.org/docs/master/torch.html?highlight=unbind#indexing-slicing-joining-mutating-ops) to prevent future issues

lapid92:
`torch.unbind` was already in all of the PyTorch TPCs except tflite, which I added

Idan-BenAmi (approved):

elad-c (approved):

PR #765, opened 2023-08-15
support tf2.13

eladc-git:
TF support 2.13:
1. remove support from version < 2.6
2. add imports for TF.2.13 and above
3. Adapt serialization for new TF.2.13

### DIFF
--- a/model_compression_toolkit/trainable_infrastructure/keras/config_serialization.py
+++ b/model_compression_toolkit/trainable_infrastructure/keras/config_serialization.py
@@ -17,11 +17,16 @@ import copy
 from typing import Any, Union
 from enum import Enum
 
+import numpy as np
+
 from model_compression_toolkit.target_platform_capabilities.target_platform import QuantizationMethod
 from model_compression_toolkit.trainable_infrastructure.common.trainable_quantizer_config import \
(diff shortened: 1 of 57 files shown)


model_compression_toolkit/trainable_infrastructure/keras/config_serialization.py
```diff
@@ -17,12 +17,17 @@
...
 
+import numpy as np
+
 from model_compression_toolkit.target_platform_capabilities.target_platform import QuantizationMethod
 from model_compression_toolkit.trainable_infrastructure.common.trainable_quantizer_config import \
     TrainableQuantizerActivationConfig, TrainableQuantizerWeightsConfig
 from mct_quantizers.common import constants as C
 
 
+CONFIG = "config"
```
ofirgo, line 28: maybe put these in our constants file? (and then use them in relevant exporter tests as well)

ofirgo (approved):

model_compression_toolkit/trainable_infrastructure/keras/config_serialization.py
eladc-git: I thought about that but is very very specific to here

PR #770, opened 2023-08-15
Refactor workflows

reuvenperetz:
1. Creating two basic workflows: one for pytorch and one for keras.
2. Creating workflows that use the basic workflows for each possible combination of a nn-framework version and a python version (we don't use matrix strategy since a badge is needed for each combination for the readme.
3. Creating one workflow for coverage which uses the latest supported nn-frameworks versions.
4. Update README file.


.github/workflows/run_tests_python310_keras211.yml
```diff
@@ -1,14 +1,14 @@
...
       - main
 
+
 jobs:
-  run-tensorflow-2_12:
+  run-tests:
     uses: ./.github/workflows/run_keras_tests.yml
     with:
-      tf-version: "2.12.*"
+      python-version: "3.10"
```
ofirgo, line 13: maybe use 3.10.* ? I'm not sure, but I remember that I encountered an issue when using 3.10 because sometimes it ignores the "0" and refers to it as "3.1"

.github/workflows/run_tests_python310_keras212.yml
```diff
@@ -1,14 +1,14 @@
...
       - main
 
+
 jobs:
-  run-tensorflow-2_13:
+  run-tests:
     uses: ./.github/workflows/run_keras_tests.yml
     with:
-      tf-version: "2.13.*"
+      python-version: "3.10"
```
ofirgo, line 13: same if you decide to change

.github/workflows/run_tests_python310_keras213.yml
```diff
@@ -0,0 +1,14 @@
...
+  pull_request:
+    branches:
+      - main
+
+
+jobs:
+  run-tests:
+    uses: ./.github/workflows/run_keras_tests.yml
+    with:
+      python-version: "3.10"
```
ofirgo, line 13: same

ofirgo (approved):

PR #774, opened 2023-08-16
add .keras format saving to tests

eladc-git:


model_compression_toolkit/exporter/model_exporter/keras/fakely_quant_tflite_exporter.py
```diff
@@ -18,6 +18,11 @@
 
 import keras.models
 import tensorflow as tf
+from packaging import version
+if version.parse(tf.__version__) >= version.parse("2.13"):
+    SAVE_FORMAT = ".keras"
+else:
+    SAVE_FORMAT = ".h5"
```
reuvenperetz, line 25: I think it's better to define this once in the base keras exporter and use it in all other parts (instead of duplicating it).
Also, since it's used only for tmp file saving (like tflite export or tests), I would consider renaming it to something like "TMP_KERAS_EXPORT_FORMAT" so we will avoid confusion in the future (since the user can pass a path with either h5 or keras extension)

model_compression_toolkit/exporter/model_exporter/keras/export_serialization_format.py
```diff
@@ -19,7 +19,7 @@ class KerasExportSerializationFormat(Enum):
     """
     Specify which serialization format to use for exporting a quantized Keras model.
 
-    KERAS_H5 - .h5 file format
+    KERAS_H5 - .keras (TF2.13 and above) or .h5 (TF2.12 and below)file format
```
reuvenperetz, line 22: Please fix missing space

Issue #771, opened 2023-08-16
SNC crashes on efficientnet_b0


mikeseven:
### Issue Type

Bug

### Source

source

### MCT Version

main

### OS Platform and Distribution

ubuntu 20.04

### Python version

3.10

### Describe the issue

```shell
shift negative correction crashes with efficientnet_b0 with pytorch because the tensors passed to torch.fake_quantize_affine are numpy arrays instead of torch tensors.
```

### Expected behaviour

```shell
no crash
```

### Code to reproduce the issue

```shell
download efficientnet_b0 from torchvision model hub
create typical 8b quantizer for both weights and activations, with apply SNC true.
```

mikeseven:
Note: this is on Pytorch 2 and calibration on imagenet.

reuvenperetz:
Hello @mikeseven,
What are you using as inputs for inference?
I reproduced your error when the inputs are numpy arrays, but it works well with torch tensors.

Can you please run this code to see if it is reproduced?
```
import torchvision
import model_compression_toolkit as mct
import numpy as np
import torch

model = torchvision.models.efficientnet_b0(pretrained=True)

def representative_data_gen():
    for i in range(1):
        yield [torch.rand(1, 3, 224, 224)]

tpc = mct.get_target_platform_capabilities('pytorch', 'default')

core_config = mct.core.CoreConfig(quantization_config=mct.core.QuantizationConfig(shift_negative_activation_correction=True))

quantized_model, quantization_info = mct.ptq.pytorch_post_training_quantization_experimental(model,
                                                                                             representative_data_gen,
                                                                                             core_config=core_config,
                                                                                             target_platform_capabilities=tpc)
outputs=quantized_model(torch.rand(1,3,224,224))

```

Best Regards

mikeseven:
Thanks for your reply and many thanks for this great toolkit.

Here is the config I used in your code:
```python
core_config = mct.core.CoreConfig(
    quantization_config=mct.core.QuantizationConfig(
        activation_error_method=mct.core.QuantizationErrorMethod.MSE,  
        weights_error_method=mct.core.QuantizationErrorMethod.NOCLIPPING, 
        weights_per_channel_threshold=True,  # per-channel weights
        relu_bound_to_power_of_2=False,  # True may provide slightly better accuracy
        weights_bias_correction=True,  # for stats correction of bias
        weights_second_moment_correction=True,  # for stats correction of variance
        input_scaling=False,  # not supported for pytorch yet
[...]

PR #778, opened 2023-08-17
Fix params search in pytorch SNC

reuvenperetz:
In SNC, when better-shifting values are tested for a better choice, the histogram (which is a numpy object) is quantized using the non-linear node activation quantization function (to estimate the expected mse compared to the original histogram). The issue is that this quantization function is a framework tensor, which makes it fail since it expects a fw tensor. To overcome the issue, this commit adds to the common part of SNC an argument which is a callable that receives two arguments and returns one: it gets the fw activation quantization function and the bins to quantize. The function (of each fw) responsible for doing (if needed) a preprocessing and postprocessing to the bins, which is a numpy object.
Only if this function is passed - common SNC will use this function. If not - it assumes no processing is needed and simply uses the activation quantization quantizer to quantize the bins (like tf, for example).


reuvenperetz:
Fix #771

eladc-git (approved):

PR #779, opened 2023-08-17
Fix missing prior info for new nodes in SNC

reuvenperetz:
Add to NodePriorInfo a method to create a shifted version of it. Use it in SNC to attach prior info for new Add nodes.


reuvenperetz:
Fix #771

Idan-BenAmi (approved):

PR #782, opened 2023-08-18
better separation keras/pytorch and pytorch set device

mikeseven:
same as for mct_quantizers:
- better separation of keras and pytorch imports in case one is missing so importing this module won't crash
- also setting pytorch device


tutorials/quick_start/results/model_quantization_results.csv
```diff
@@ -1,13 +1,13 @@
...
-mobilenet_v2,torchvision_lib,ImageNet,50000,0.7215,0.7188,3.47,W8A8,PTQ,default_pytorch_tpc-v4
-regnet_y_400mf,torchvision_lib,ImageNet,50000,0.7578,0.7542,4.31,W8A8,PTQ,default_pytorch_tpc-v4
-shufflenet_v2_x0_5,torchvision_lib,ImageNet,50000,0.6055,0.597,1.36,W8A8,PTQ,default_pytorch_tpc-v4
-squeezenet1_0,torchvision_lib,ImageNet,50000,0.581,0.5767,1.24,W8A8,PTQ,default_pytorch_tpc-v4
-regnetx_002,timm_lib,ImageNet,50000,0.6876,0.6827,2.67,W8A8,PTQ,default_pytorch_tpc-v4
-regnety_008,timm_lib,ImageNet,50000,0.7632,0.7598,6.23,W8A8,PTQ,default_pytorch_tpc-v4
-resnet10t,timm_lib,ImageNet,50000,0.6656,0.6643,5.43,W8A8,PTQ,default_pytorch_tpc-v4
-resnet18,timm_lib,ImageNet,50000,0.6976,0.6961,11.68,W8A8,PTQ,default_pytorch_tpc-v4
-efficientnet_es,timm_lib,ImageNet,50000,0.7808,0.7774,5.39,W8A8,PTQ,default_pytorch_tpc-v4
+mobilenet_v2,torchvision,ImageNet,50000,0.7215,0.7188,3.47,W8A8,PTQ,default_pytorch_tpc-v4
```
ofirgo, line 2: The 'lib' suffix is meant to avoid conflicts with the original package names.
Please remove the folder name changes.

model_compression_toolkit/exporter/model_wrapper/__init__.py
```diff
@@ -13,8 +13,12 @@
 # limitations under the License.
 # ==============================================================================
 
-from model_compression_toolkit.exporter.model_wrapper.keras.validate_layer import is_keras_layer_exportable
-from model_compression_toolkit.exporter.model_wrapper.keras.builder.fully_quantized_model_builder import get_exportable_keras_model
+from mct_quantizers.common import constants
 
-from model_compression_toolkit.exporter.model_wrapper.pytorch.validate_layer import is_pytorch_layer_exportable
-from model_compression_toolkit.exporter.model_wrapper.pytorch.builder.fully_quantized_model_builder import get_exportable_pytorch_model
+if constants.FOUND_TF:
```
ofirgo, line 18: All these API functions/classes have protection against missing tf/torch packages in the function/class level (with FOUND_TF or FOUND_TORCH).
So you don't need to add it here as well.
Did you encounter any missing FOUND_TF/FOUND_TORCH in any of the APIs? If so, it would be better to add it there.

model_compression_toolkit/core/pytorch/utils.py
```diff
@@ -17,6 +17,8 @@
 from typing import Union
 from model_compression_toolkit.core.pytorch.constants import CUDA, CPU
 
+_device = CUDA if torch.cuda.is_available() else CPU
```
ofirgo, line 20: What is the incentive for defining this global variable?
In my opinion, it is better to avoid such variables, and there is no need to enable device initialization from an outer scope.

model_compression_toolkit/exporter/model_wrapper/__init__.py
mikeseven: I had to put them because it would crash otherwise because I don't use TF.

model_compression_toolkit/core/pytorch/utils.py
mikeseven: Agree, it's not pretty but it was just a quick way to enable multi gpu support across Mct.

tutorials/quick_start/results/model_quantization_results.csv
mikeseven: The reason I did that is because it's inconsistent between code and example. And the code didn't work. It's obvious to me that this indicates where the model comes from, not the package name. There is no need for lib suffix even in the code because of the way the packages are loaded; no conflict possible. 
Either way, with or without lib suffix, is fine with me.

ofirgo: @Idan-BenAmi 
Please take a look

model_compression_toolkit/exporter/model_wrapper/__init__.py
ofirgo: > I had to put them because it would crash otherwise because I don't use TF.

Can you detail in which file the error occurs? Can you provide the error stack?
It might be that there is a missing FOUND_TF condition somewhere else in the code.

model_compression_toolkit/core/pytorch/utils.py
ofirgo: @haihabi 
Please take a look

ofirgo: Hi @mikeseven,

Can you please modify this change to not use "global" in order to update the _device variable?
Besides that, I agree that this modification is useful.

model_compression_toolkit/exporter/model_wrapper/__init__.py
ofirgo: Hi @mikeseven,

Can you please provide the information regarding where the program is failing because of this missing condition so we can validate it and move forward with the PR?

Thanks

tutorials/quick_start/results/model_quantization_results.csv
Idan-BenAmi: Hi,
Thanks for this comment! We already added a fix that resolved this naming problem here: https://github.com/sony/model_optimization/pull/785
Therefore this change is redundant now.

PR #833, opened 2023-10-17
Hessian-based scores w.r.t weights (Keras)

reuvenperetz:
Add implementation for Hessian-based scores w.r.t node's weight in Keras.
Add the support of HessianInfoService to receive requests when the mode is 'weights'.

Several general changes:
- Fix in fetch_hessians for the case of zero requested approximations.
- Move outputs concatenation (for cases the model has multiple outputs) to framework (keras/pytorch) calculators.
- Fix missing licenses in multiple files.
- Fixed issue in separable decomposition regarding the reuse group name (found due to assertions in reuse case in the HessianServiceInfo).


model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
```diff
@@ -50,5 +53,112 @@ def __init__(self,
...
+        shape of (3, 3, 3, 2) (namely, 3 input channels, 2 output channels and kernel size of 3x3)
+        and the required granularity is HessianInfoGranularity.PER_TENSOR the result shape will be (1,),
+        for HessianInfoGranularity.PER_OUTPUT_CHANNEL the shape will be (2,) and for
+        HessianInfoGranularity.PER_ELEMENT a shape of (3, 3, 3, 2).
+
+        Returns:  The computed scores as numpy ndarray.
+
+        """
+        # Check if the target node's layer type is supported
+        if self.hessian_request.target_node.layer_class not in [Conv2D, Dense, Conv2DTranspose, DepthwiseConv2D]:
```
eladc-git, line 71: Why not using the layers with kernels defined in DEFAULT_KERAS_INFO instead of explicitly writing them? This is useful if tomorrow I will add Conv1D, I will update only DEFAULT_KERAS_INFO  and don't need to add it also here.

model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
```diff
@@ -50,5 +53,112 @@ def __init__(self,
...
+            Logger.error(
+                f"{self.hessian_request.target_node.type} is not supported for Hessian info w.r.t weights.")
+
+        # Construct the Keras float model for inference
+        model, _ = FloatKerasModelBuilder(graph=self.graph).build_model()
+        # TODO: what to do in reuse
```
eladc-git, line 77: what is the problem in reuse?

model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
```diff
@@ -50,5 +53,112 @@ def __init__(self,
...
+
+        # Get the output channel index (needed for HessianInfoGranularity.PER_OUTPUT_CHANNEL case)
+        output_channel_axis, _ = DEFAULT_KERAS_INFO.kernel_channels_mapping.get(
+            self.hessian_request.target_node.type)
+
+        # Initiate a gradient tape for automatic differentiation
+        with tf.GradientTape(persistent=True) as tape:
+            # Perform a forward pass (inference) to get the output, while watching
+            # the input tensor for gradient computation
+            tape.watch(self.input_images)
```
eladc-git, line 94: I know this was already done in previous PR but It is not smooth when the input is called input_images. (what if it is not images). input_tensors is more accurate.

model_compression_toolkit/core/keras/hessian/trace_hessian_calculator_keras.py
```diff
@@ -48,4 +56,25 @@ def __init__(self,
                                                           trace_hessian_request=trace_hessian_request,
                                                           num_iterations_for_approximation=num_iterations_for_approximation)
 
+    def _concat_outputs(self, outputs: Union[tf.Tensor, List[tf.Tensor]]) -> tf.Tensor:
```
eladc-git, line 59: same here. Create a general concat function (one use is for concat multiple model's output).

model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
```diff
@@ -50,5 +53,112 @@ def __init__(self,
...
+                v = tf.random.normal(shape=output.shape)
+                f_v = tf.reduce_sum(v * output)
+
+                # Stop recording operations for automatic differentiation
+                with tape.stop_recording():
+                    # Compute gradients of f_v with respect to the weights
+                    gradients = tape.gradient(f_v, weight_tensor)
+
+                    # Reshape the gradients based on the granularity (whole tensor, per channel, or per element)
+                    num_of_scores = self._get_num_scores_by_granularity(gradients,
```
eladc-git, line 112: this function should be above the grad calculation and gets weights and not gradiets. You don't need to calc it each iteration.

model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
```diff
@@ -50,5 +53,112 @@ def __init__(self,
...
+        for HessianInfoGranularity.PER_OUTPUT_CHANNEL the shape will be (2,) and for
+        HessianInfoGranularity.PER_ELEMENT a shape of (3, 3, 3, 2).
+
+        Returns:  The computed scores as numpy ndarray.
+
+        """
+        # Check if the target node's layer type is supported
+        if self.hessian_request.target_node.layer_class not in [Conv2D, Dense, Conv2DTranspose, DepthwiseConv2D]:
+            Logger.error(
+                f"{self.hessian_request.target_node.type} is not supported for Hessian info w.r.t weights.")
```
eladc-git, line 73: I wonder if we want to return error for every layer we don't support like you did or just for layer with weights we don't support. For now it means someone should walk over all layers and pick only the ones with weights that we support in order no to get an error (like you did in the tests). What if when there is no weight (like relu) return None?

eladc-git (changes requested):

model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
reuvenperetz: I think that this might lead to parts of the code frequently checking for None values, which could make the code hard to read and maintain. I think it's reasonable that unsupported requests will trigger an error, ensuring clarity of usage and providing direct feedback about the exact problem.

model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
reuvenperetz: Since it's in other existing parts (like in w.r.t activations), I'll rename it in all those parts in a separate PR.

model_compression_toolkit/core/keras/hessian/weights_trace_hessian_calculator_keras.py
reuvenperetz: Since some nodes are the same layer (in the framework model), computing their scores is needed only for one of them (the scores for the rest of them will be the same scores as the one we computed once). For this reason, in the HessianInfoService we'll need to add this logic that if a request contains a reused node, we'll take the computation of the "master" node (in the reuse group).

model_compression_toolkit/core/common/hessian/trace_hessian_calculator.py
```diff
@@ -75,3 +75,24 @@ def compute(self) -> List[float]:
         computation method for the trace Hessian approximation.
         """
         raise NotImplemented(f'{self.__class__.__name__} have to implement compute method.')  # pragma: no cover
+
+    def _unfold_outputs(self, outputs: Any) -> List[Any]:
```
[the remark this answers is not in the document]
reuvenperetz: Where would you think it should be? Since it's may use other parts outside the hessian package, and we don't have a utils module (if I'm not wrong)

model_compression_toolkit/core/keras/hessian/trace_hessian_calculator_keras.py
reuvenperetz: Where would you think it should be? Since other parts outside the hessian package may use it, and we don't have a utils module (if I'm not wrong)

model_compression_toolkit/core/common/hessian/trace_hessian_calculator.py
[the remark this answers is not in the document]
eladc-git: I meant the name and the variables in this function, not the location. Because its functionality is unfolding: taking tensor and return list the using the names "output" in this function is not neccecry. For example the input can be "input_tensor" instead of "outputs".

eladc-git (approved):

PR #834, opened 2023-10-18
Fix issue 824

elad-c:
This PR allows using Keras custom layers in the Keras model. A custom layers can be added by defining it in an OperatorSet in the TPC.


model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -91,6 +91,13 @@ def set_tpc(self,
         Args:
             tpc: TargetPlatformCapabilities object.
         """
+        # validate graph nodes are either from the framework or a custom layer defined in the TPC
+        tpc_layers = tpc.op_sets_to_layers.get_layers()
+        for n in self.nodes:
+            if n.is_custom and n.type not in tpc_layers:
```
reuvenperetz, line 97: We need to match also LayerFilterParams to see if they were defined in the TPC

reuvenperetz: There's a match function that may help you with it for this case

reuvenperetz:
Maybe we should raise an error if it's configured in the TPC that this op's weights should be quantized?

reuvenperetz (approved):

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -467,6 +469,10 @@ def is_match_filter_params(self, layer_filter_params: LayerFilterParams) -> bool
         Returns:
             Whether the node matches to the LayerFilterParams properties.
         """
+        # check if passed argument is of type LayerFilterParams
```
Idan-BenAmi, line 472: # Check if the provided argument ...

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -91,6 +92,18 @@ def set_tpc(self,
         Args:
             tpc: TargetPlatformCapabilities object.
         """
+        # validate graph nodes are either from the framework or a custom layer defined in the TPC
```
Idan-BenAmi, line 95: --> # Validate graph nodes are either built-in layers from the framework or custom layers defined in the TPC

Idan-BenAmi (approved):
Few small comments

PR #837, opened 2023-10-22
Add testing using TF2.14

reuvenperetz:
Replace workflows that use tf2.11 with tf2.14


eladc-git (commented):
What about TF2.14 in the main readme?

eladc-git (approved):

PR #840, opened 2023-10-23
Fix allocation memory in Keras tests

reuvenperetz:


tests/keras_tests/non_parallel_tests/test_tensorboard_writer.py
```diff
@@ -139,7 +139,8 @@ def test_steps_by_order(self):
         def rep_data():
             yield [np.random.randn(1, 8, 8, 3)]
 
-        mp_qc = mct.core.MixedPrecisionQuantizationConfigV2(num_of_images=1)
+        mp_qc = mct.core.MixedPrecisionQuantizationConfigV2(num_of_images=1,
+                                                            use_grad_based_weights=False)
```
Idan-BenAmi, line 143: why use_grad_based_weights=False if it's a non parallel test?

Idan-BenAmi (approved):

tests/keras_tests/non_parallel_tests/test_tensorboard_writer.py
reuvenperetz: To speed it up. The tensorboard  test does not test anything regarding the gradients weighting.

PR #843, opened 2023-10-29
Fix mixed precision LP numerical issue

ofirgo:
Adding a threshold for mixed precision metric and scaling the metric in case some values are larger than the threshold, to prevent numerical issues.
This is supposed to address a possible issue with the pulp library's default LP solver.

### DIFF
--- a/model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
+++ b/model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
@@ -64,7 +64,7 @@ class MixedPrecisionSearchManager:
         self.fw_impl = fw_impl
         self.sensitivity_evaluator = sensitivity_evaluator
         self.layer_to_bitwidth_mapping = self.get_search_space()
-        self.sensitivity_metric_fn = self.get_sensitivity_metric()
+        self.compute_metric_fn = self.get_sensitivity_metric()
 
         self.compute_kpi_functions = kpi_functions
         self.target_kpi = target_kpi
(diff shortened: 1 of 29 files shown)


model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
```diff
@@ -289,6 +289,30 @@ def compute_kpi_for_config(self, config: List[int]) -> KPI:
...
+
+        """
+        # normalize metric for numerical stability
+
+        max_dist = max([max([d for b, d in dists.items()]) for layer, dists in layer_to_metrics_mapping.items()])
+        if max_dist >= self.sensitivity_evaluator.quant_config.metric_normalization_threshold:
+            Logger.warning(f"The mixed precision distance metric values indicate a large error in the quantized model."
+                           f"this can cause numerical issues."
+                           f"The program will proceed with mixed precision search after scaling the metric values,"
+                           f"which can lead to unstable results.")
```
Idan-BenAmi, line 311: Wouldn't it be more correct? The high values in the mixed precision distance metric may imply a large error in the quantized model which can cause numerical issues

Idan-BenAmi (approved):

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
ofirgo: It is to inform the user that there might be an issue with the model, not the mixed precision necessarily. Such large distance may indicates that something is wrong with the model.

Issue #841, opened 2023-10-24
torch.save fails with pickling error on mct quantized model 


irenaby:
### Issue Type

Bug

### Source

pip (model-compression-toolkit)

### MCT Version

1.10.0

### OS Platform and Distribution

Ubuntu 20.04

### Python version

3.10

### Describe the issue

```shell
torch.save fails with pickling error on mct quantized model
```

### Code to reproduce the issue

```shell
import torch
import torchvision as tv
import model_compression_toolkit as mct

model = tv.models.get_model('alexnet')
model.eval()
def gen():
    yield [torch.rand(1, 3, 224, 224)]
qmodel, qinfo = mct.ptq.pytorch_post_training_quantization_experimental(model, gen)
torch.save(qmodel, '/tmp/qmodel.pt')
```

### Log output

```shell
Traceback (most recent call last):
  File "/Vols/vol_design/tools/swat/users/irenab/ml-frameworks-uni-converter/uni/fx_rename/fx_parser/tests/mct_bug.py", line 37, in <module>
    torch.save(qmodel, '/tmp/qmodel.pt')
  File "/data/projects/swat/envs/irenab/py3.10_torch2.0/lib/python3.10/site-packages/torch/serialization.py", line 441, in save
    _save(obj, opened_zipfile, pickle_module, pickle_protocol)
  File "/data/projects/swat/envs/irenab/py3.10_torch2.0/lib/python3.10/site-packages/torch/serialization.py", line 653, in _save
    pickler.dump(obj)
_pickle.PicklingError: Can't pickle <function <lambda> at 0x7fb0557251b0>: attribute lookup <lambda> on model_compression_toolkit.core.pytorch.default_framework_info failed
```

PR #861, opened 2023-11-20
Fix issue #841: fix MCT so quantized models work with torch.save

elad-c:
## Pull Request Description:
Fix FrameWorkInfo DefaultDict so MCT quantized model will work with torch.save

### DIFF
--- a/model_compression_toolkit/core/common/defaultdict.py
+++ b/model_compression_toolkit/core/common/defaultdict.py
@@ -14,27 +14,28 @@
 # ==============================================================================
 
 
-from typing import Callable, Dict, Any
+from typing import Dict, Any
+from copy import deepcopy
 
 
-class DefaultDict(object):
+class DefaultDict:
     """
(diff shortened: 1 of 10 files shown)


model_compression_toolkit/core/common/defaultdict.py
```diff
@@ -14,27 +14,28 @@
...
     Default dictionary. It wraps a dictionary given at initialization and return its
     values when requested. If the requested key is not presented at initial dictionary,
-    it returns the returned value a default factory (that is passed at initialization) generates.
+    it returns the returned value a default value (that is passed at initialization) generates.
     """
 
     def __init__(self,
                  known_dict: Dict[Any, Any],
-                 default_factory: Callable = None):
+                 default_value: Any = None):
```
eladc-git, line 30: why we need known_dict and default_value? I see we use only one of them each instance and it makes the init to be with empy dictionary. You can use a single argument and check inside if it dictionary or not.

eladc-git (approved):

PR #867, opened 2023-11-22
Fix bug in pytorch reshape substitute

eladc-git:


model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/reshape_with_static_shapes.py
```diff
@@ -50,6 +50,10 @@ def substitute(self,
         Returns:
             Graph after applying the substitution.
         """
+        # skip substitution if the reshape is applying on weights (and not activations) so the node is first source and doesn't have input shape
+        if len(node.input_shape) == 0:
+            return graph
+
         # we want the batch size value to infer from the length of the array and remaining dimensions
```
lior-dikstein, line 57: # Skip substitution if the reshape is applied to weights (not activations); 
# in this case, the node is the first source and doesn't have an input shape.

lior-dikstein (approved):

PR #894, opened 2023-12-26
Create Softmax layer in activation decomposition instead of Activation layer

reuvenperetz:
## Pull Request Description:

In MP we need to know the axis used in Softmax for KL-Divergence computation. In activation decomposition, we created a simple 'Activation' layer thus we can not use the axis info when computing MP distance. This commit changes the decomposition to create a Softmax layer, thus enabling the KL-Divergence in MP distance computation.

### DIFF
--- a/model_compression_toolkit/core/keras/graph_substitutions/substitutions/activation_decomposition.py
+++ b/model_compression_toolkit/core/keras/graph_substitutions/substitutions/activation_decomposition.py
@@ -71,8 +71,7 @@ class ActivationDecomposition(common.BaseSubstitution):
         activation_node_name = op2d_node.name + '_post_activation'
 
         # Softmax is a special case where we need to know the default axis parameter used
-        # (for the distance metric in MP for KL-Divergence), and for this reason we create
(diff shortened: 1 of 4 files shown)


model_compression_toolkit/core/keras/graph_substitutions/substitutions/activation_decomposition.py
```diff
@@ -70,18 +70,30 @@ def substitute(self,
...
-            ACTIVATION: op2d_node.framework_attr.get(ACTIVATION)}
-
-        activation_node = common.graph.BaseNode(activation_node_name,
-                                                activation_fw_attr,
-                                                op2d_node.output_shape,
-                                                op2d_node.output_shape,
-                                                {},
-                                                Activation)
+        # Softmax is a special case where we need to know the default axis parameter used
+        # (for the distance metric in MP for KL-Divergence), and for this reason we create
```
ofirgo, line 74: I would remove the cause reference (MP and KL) from here, it can be confusing if you just looked at this file without context.

ofirgo (approved):

PR #964, opened 2024-02-29
Support multiple weights attributes quantization in core

ofirgo:
## Pull Request Description:

This PR adds a support in multiple attributes quantization from the MCT core side.
There are 3 main parts to this change:
- Modifying the node candidate quantization configurations list to include per-attribute candidates: a new class WeightsAttrQuantizationConfig was added and a config of this type is created for each config that appears in the attribute's TPC configs mapping.
Note that most of the code in this class was moved from the class NodeWeightsQuantizationConfig, because now most of the node's quantization parameters are per-attribute and not per-node.
- Enabling per-attribute operations - quantization parameters search and quantization are now done per-attribute, based on the attribute config, instead of on the kernel attribute alone.
- Restricting several features for kernel-only:
- - Mixed precision, GPTQ, QAT, and pruning currently continue to work only on the kernel attribute. Several TODOs are left in the code in places we need to address later.
- - Several features, like SNC, bias correction, etc. are relevant only for nodes that have a quantized kernel, therefore, in those places we needed to add support in extracting the kernel attribute from the node (using the fw info mapping).

Important: since currently our default TPC do not enables quantization of other attributes beside the kernel, no change in behavior is expected after this PR. Nevertheless, users will be able to create their own TPC with other attributes quantization - this feature should work, but is defined as experimental at the moment (see validation and warning in set_node_quantization_config.py)

FYI, wherever there is a missing typehint for fw_info or fw_implementation - this is to avoid a circular dependency.


model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -342,6 +388,21 @@ def has_weights_to_quantize(self, fw_info):
...
 
+    def has_any_weight_attr_to_quantize(self) -> bool:
+        """
+        Checks whether the node has any weights attribute that is supposed to be quantized, based on its provided
+        quantization configuration candidates.
+
+        Returns: Tru if the is at least one weights attribute in the node that is supposed to be quantized.
+
+        """
+        for attr in self.get_node_weights_attributes():
```
reuvenperetz, line 399: Not crucial, but this can be written in a single line.

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -445,15 +515,24 @@ def get_unique_activation_candidates(self) -> List[Any]:
...
+            fw_info: Framework info to decide which attributes should be quantized.
 
         Returns: True if the node has at list one quantization configuration candidate with weights quantization enabled.
         """
+        # This method is only used for Mixed precision purposes
+        kernel_attr = fw_info.get_kernel_op_attributes(self.type)
+        if kernel_attr[0] is not None:
+            kernel_attr = kernel_attr[0]
+        else:
+            return False
```
reuvenperetz, line 532: This function is a bit confusing since it can return a boolean or the kernel string... Why not return None? (I know it's not ideal as well, but I think it's a bit better)

model_compression_toolkit/core/common/mixed_precision/configurable_quantizer_utils.py
```diff
@@ -20,29 +20,44 @@
...
+        for candidate_bits in n_candidate_bits:
+            assert candidate_bits < curmax, f"Node's quantization configuration candidates should arrive in " \
+                                            f"descending order of (weights_nbits, activation_nbits)."
+            curmax = candidate_bits
+    else:
+        # The candidates are only activation configurable
+        curmax = np.inf
+        n_candidate_bits = [x.activation_quantization_cfg.activation_n_bits for x in node_q_cfg]
+        for candidate_bits in n_candidate_bits:
+            assert candidate_bits < curmax, f"Node's quantization configuration candidates should arrive in " \
```
reuvenperetz, line 52: Logger?

model_compression_toolkit/core/common/mixed_precision/configurable_quantizer_utils.py
```diff
@@ -20,29 +20,44 @@
     CandidateNodeQuantizationConfig
 
 
-def verify_candidates_descending_order(node_q_cfg: List[CandidateNodeQuantizationConfig]):
+def verify_candidates_descending_order(node_q_cfg: List[CandidateNodeQuantizationConfig],
+                                       kernel_attr: str = None):
```
reuvenperetz, line 24: Missing type hint

tests/keras_tests/feature_networks_tests/feature_networks/const_representation_test.py
```diff
@@ -30,8 +30,10 @@ class ConstRepresentationTest(BaseKerasFeatureNetworkTest):
 
     def __init__(self, unit_test, layer, const, is_list_input=False, input_reverse_order=False, use_kwrags=False,
                  input_shape=(32, 32, 16)):
+        # TODO: Running test without MCTQ exporter since this feature is not yet supported in MCTQ.
+        #  Need to change the tests once support is added.
         super(ConstRepresentationTest, self).__init__(unit_test=unit_test, input_shape=input_shape,
-                                                      experimental_exporter=True)
+                                                      experimental_exporter=False)
```
reuvenperetz, line 36: Why are these changes?

model_compression_toolkit/exporter/model_wrapper/keras/builder/fully_quantized_model_builder.py
```diff
@@ -39,7 +39,7 @@ def _get_wrapper(node: common.BaseNode,
         Returns: Wrapped layer with weights quantizers and activation quantizers
 
         """
-        weights_quantizers, _ = get_quantization_quantizers(node)
+        weights_quantizers, _ = C.keras.keras_implementation.KerasImplementation().get_quantization_quantizers(node)
```
elad-c, line 42: Can you put the `get_quantization_quantizers` in the node when craeting it, so you don't need to craete the keras implementation each time?

model_compression_toolkit/exporter/model_wrapper/pytorch/builder/fully_quantized_model_builder.py
```diff
@@ -40,7 +38,7 @@ def fully_quantized_wrapper(node: common.BaseNode,
         Returns: Wrapped layer
 
         """
-        weight_quantizers, _ = get_quantization_quantizers(node)
+        weight_quantizers, _ = C.pytorch.pytorch_implementation.PytorchImplementation().get_quantization_quantizers(node)
```
elad-c, line 41: same comment as keras

model_compression_toolkit/gptq/keras/gptq_training.py
```diff
@@ -149,11 +146,16 @@ def gptq_wrapper(self,
...
-                                                         self.gptq_config) # TODO: split quantizers building into two functions: for weights and activations
+                                                         self.gptq_config,  # TODO: split quantizers building into two functions: for weights and activations
+                                                         self.fw_info.get_kernel_op_attributes(n.type)[0])
             if len(weights_quantizers) > 0:
                 return KerasTrainableQuantizationWrapper(layer,
-                                                   weights_quantizers=weights_quantizers)
+                                                         weights_quantizers=weights_quantizers)
+
+        # TODO: need to check if in this case, if there are other weights attributes that are not trainable but are
+        #  quantized, do we need to wrap them as well?
```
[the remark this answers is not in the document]
ofirgo: You are right, this is just something that we decided not to decide on right now to prevent potential bugs and behavior changes.
Once we enable attribute quantization by default we will address this (here and also in MP & QAT)

model_compression_toolkit/trainable_infrastructure/common/get_quantizer_config.py
```diff
@@ -73,41 +76,44 @@ def get_trainable_quantizer_activation_config(
...
-    # all candidates must have the same weights quantization method
-    weights_quantization_methods = set([cfg.weights_quantization_cfg.weights_quantization_method for cfg in n.candidates_quantization_cfg])
-    if len(weights_quantization_methods) > 1:
-        Logger.error(f'Unsupported candidates_quantization_cfg with different weights quantization methods: {weights_quantization_methods}') # pragma: no cover
+
+    if attr is not None:
+        # all candidates must have the same weights quantization method
+        weights_quantization_methods = set([cfg.weights_quantization_cfg.get_attr_config(attr).weights_quantization_method
+                                            for cfg in n.candidates_quantization_cfg])
+        if len(weights_quantization_methods) > 1:
```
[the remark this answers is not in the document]
ofirgo: It can be zero (no candidates are defined for this attribute). In this case we wouldn't want to fail, just return an empty list

model_compression_toolkit/trainable_infrastructure/common/get_quantizer_config.py
```diff
@@ -73,41 +76,44 @@ def get_trainable_quantizer_activation_config(
...
+            Logger.error(f'Unsupported candidates_quantization_cfg with different weights quantization methods: '
+                         f'{weights_quantization_methods}')  # pragma: no cover
 
     # all candidates must have the same activation quantization method
-    activation_quantization_methods = set([cfg.activation_quantization_cfg.activation_quantization_method for cfg in n.candidates_quantization_cfg])
+    activation_quantization_methods = set([cfg.activation_quantization_cfg.activation_quantization_method
+                                           for cfg in n.candidates_quantization_cfg])
     if len(activation_quantization_methods) > 1:
-        Logger.error(f'Unsupported candidates_quantization_cfg with different activation quantization methods: {activation_quantization_methods}') # pragma: no cover
+        Logger.error(f'Unsupported candidates_quantization_cfg with different activation quantization methods: '
```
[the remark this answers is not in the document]
ofirgo: same answer

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -445,15 +515,24 @@ def get_unique_activation_candidates(self) -> List[Any]:
                              and not seen_candidates.add(candidate.activation_quantization_cfg)]
         return unique_candidates
 
-    def has_weights_quantization_enabled_candidate(self) -> bool:
+    def has_kernel_quantization_enabled_candidate(self, fw_info) -> bool:
```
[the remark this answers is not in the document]
ofirgo: I don't want it to be more general, this is now a kernel-specific method

model_compression_toolkit/core/common/graph/base_node.py
ofirgo: Removing the function, we are only using it in one place, so I just implement the validation there, it shouldn't be part of BaseNode

model_compression_toolkit/core/common/mixed_precision/configurable_quantizer_utils.py
ofirgo: It's more of an inner validation that the user doesn't have anything to do about (he shouldn't even know about that), so I'll keep it just an assertion for now (it was an assertion before it was just copied from somewhere else so it seems new code but it's not)

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -185,6 +195,14 @@ def get_weights_list(self):
         return [self.weights[k] for k in self.weights.keys()
                 if self.weights[k] is not None and not isinstance(k, int)]
 
+    def get_node_weights_attributes(self) -> List[str]:
+        """
+
+        Returns: A list of all weights attributes that the node holds.
+
+        """
+        return [k for k in self.weights.keys() if self.weights[k] is not None]
```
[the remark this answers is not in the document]
ofirgo: let's talk about (2) offline

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -540,3 +633,27 @@ def get_simd(self) -> int:
...
+            the candidates in descending order.
+        The operation is done inplace.
+
+        Args:
+            fw_info: FrameworkInfo object about the specific framework (e.g., attributes of different layers' weights to quantize).
+
+        """
+        if self.candidates_quantization_cfg is not None:
+            kernel_attr = fw_info.get_kernel_op_attributes(self.type)[0]
+            if kernel_attr is not None:
```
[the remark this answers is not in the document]
ofirgo: I know that it's considered bad practice to [assign lambda into a variable](https://peps.python.org/pep-0008/#programming-recommendations).
Did you mean that it should be in a nested function instead of lambda? maybe but then I'll need to different nested functions and I don't think it's neater.

model_compression_toolkit/core/common/mixed_precision/bit_width_setter.py
```diff
@@ -36,35 +36,40 @@ def set_bit_widths(mixed_precision_enable: bool,
...
+        for node in graph.nodes:  # set a specific node qc for each node final qc
             # If it's reused, take the configuration that the base node has
             node_name = node.name if not node.reuse else '_'.join(node.name.split('_')[:-2])
             if node_name in sorted_nodes_names:  # only configurable nodes are in this list
                 node_index_in_graph = sorted_nodes_names.index(node_name)
                 _set_node_final_qc(bit_widths_config,
                                    node,
-                                   node_index_in_graph)
+                                   node_index_in_graph,
+                                   graph.fw_info)
```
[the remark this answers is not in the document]
ofirgo: I thought about it, but I don't want the node to have a specific variable that is dedicated to the kernel, since it is an abstraction of any general op that is not necessarily aware of what kernel is.
We can discuss this offline, but I think that since for now our methodology in the entire code regarding the kernel specifications is done using the fw info, I prefer to keep it that way

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -383,10 +375,204 @@ def __hash__(self):
...
+        self.weights_bias_correction = qc.weights_bias_correction
+
+        # Initialize a quantization configuration for each of the node's attributes
+        self.attributes_config_mapping = {}
+        self.pos_attributes_config_mapping = {}
+        for attr in node_attrs_list:
+            if isinstance(attr, int):
+                # this is a positional attribute, so it needs to be handled separately.
+                # we assume that a positional attribute is quantized with the default configuration provided in the TPC.
+                self.pos_attributes_config_mapping[attr] = WeightsAttrQuantizationConfig(qc=qc,
```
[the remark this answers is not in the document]
ofirgo: You mean just not separate the WeightsAttrQuantizationConfig creation?
See the modified code and if it's not what you meant then let's take this offline

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -383,10 +375,204 @@ def __hash__(self):
...
+            node_attrs_list: A list of the node's weights attributes names.
+
+        """
+        self.min_threshold = qc.min_threshold
+        self.simd_size = op_cfg.simd_size
+        self.weights_second_moment_correction = qc.weights_second_moment_correction
+        self.weights_bias_correction = qc.weights_bias_correction
+
+        # Initialize a quantization configuration for each of the node's attributes
+        self.attributes_config_mapping = {}
```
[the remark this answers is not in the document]
ofirgo: I think that it is simpler to maintain that way, I prefer not to have a unified dictionary with two different types of keys (string and int).
I see the positional weights "feature" as a separate capability that we need to maintain carefully, and I think that trying to be "generic" here would raise difficulties in the future.
If you think that it is much better to use a unified dict instead, let's discuss this.

model_compression_toolkit/core/keras/keras_implementation.py
```diff
@@ -586,3 +590,38 @@ def sensitivity_eval_inference(self,
...
+        Args:
+           node: Node to get quantizers for.
+
+        Returns:
+            weight_quantizers: A dictionary between a weight's name to its quantizer.
+            activation_quantizers: A list of activations quantization, one for each layer output.
+
+        """
+
+        def _weight_name(w: str) -> str:
```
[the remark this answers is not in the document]
ofirgo: It is a protected class member of the wrapper in MCTQ so I don't want to import it here (not sure that I can even if I wanted to)

model_compression_toolkit/exporter/model_wrapper/keras/builder/fully_quantized_model_builder.py
ofirgo: Not sure that I understand your suggestion, let's talk about it offline

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -383,10 +375,206 @@ def __hash__(self):
...
+            self.weights_bias_correction == other.weights_bias_correction and \
+            self.attributes_config_mapping.keys() == other.attributes_config_mapping.keys() and \
+            all([self.attributes_config_mapping[k] == other.attributes_config_mapping[k]
+                 for k in self.attributes_config_mapping.keys()]) and \
+            self.pos_attributes_config_mapping.keys() == other.pos_attributes_config_mapping.keys() and \
+            all([self.pos_attributes_config_mapping[k] == other.pos_attributes_config_mapping[k]
+                 for k in self.pos_attributes_config_mapping.keys()])
+
+    def __hash__(self):
+        return hash((self.min_threshold,
```
reuvenperetz, line 576: Missing weights_bias_correction?

model_compression_toolkit/trainable_infrastructure/common/get_quantizer_config.py
```diff
@@ -73,41 +76,44 @@ def get_trainable_quantizer_activation_config(
                                               activation_quantization_candidates)
 
 
-def get_trainable_quantizer_quantization_candidates(n: BaseNode):
+def get_trainable_quantizer_quantization_candidates(n: BaseNode, attr: str):
```
reuvenperetz, line 79: Maybe add default None for the activation case?

reuvenperetz (approved):

PR #990, opened 2024-03-11
Move MP traget_kpi from facade to MixedPrecisionQuantizationConfig

ofirgo:
## Pull Request Description:

This change solves two problems:
-
- Usability: now it's clear that in order to enable mixed precision quantization, the user should provide a mixed precision configuration inside the core config (we still check that the target KPI is not None, but this is not part of "mixed_precision_enabled" condition).
- 
- API clarity: target_kpi is not a part of PQT, GPTQ of QAT facades, and it was unreasonable to include it there, since it is only relevant for mixed precision (running target_kpi remains a part of running facade arguments).


elad-c (approved):
It's about time

PR #1042, opened 2024-04-14
Hessian MSE for weights parameters search

ofirgo:
## Pull Request Description:

Implementing Hessian MSE (HMSE) for weights quantization parameters selection, using per-element Hessian-based scores.
This feature is currently available only when running the program as part of a GPTQ optimization, because of the long runtime of the Hessian computation.
It is also only available for kernel attribute parameter selection at the moment.


model_compression_toolkit/core/common/quantization/quantization_params_generation/error_functions.py
```diff
@@ -371,13 +374,62 @@ def _get_sliced_histogram(bins: np.ndarray,
     return bins_subset, counts_subset
 
 
+def _compute_hessian_for_hmse(node, hessian_info_service: HessianInfoService,
```
reuvenperetz, line 377: The type hint is missing due to import issues?
Also, start each argument in a new line, please.

reuvenperetz (approved):

model_compression_toolkit/constants.py
```diff
@@ -97,6 +97,8 @@
 DEC_RANGE_BOTTOM = 0.97
 DEC_RANGE_UPPER = 1.03
 
+NUM_QPARAM_HESSIAN_SAMPLES = 16
+
```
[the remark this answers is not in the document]
ofirgo: I thought about it, I think that for now we'll keep it a constant and not let the user control it, it is too "implementation details-related" in my opinion.

model_compression_toolkit/core/common/quantization/quantization_params_generation/error_functions.py
ofirgo: Yes, there were several places where I couldn't import BaseNode without getting into circular dependency issues.

PR #1045, opened 2024-04-17
Const quantization

elad-c:
## Pull Request Description:
This PR enables quantizing constants in the model. Constants inputs to `add`, `sub`, `mul` & `div` will be quantized when using the IMX500.v2 TPC.

### DIFF
--- a/model_compression_toolkit/core/common/graph/functional_node.py
+++ b/model_compression_toolkit/core/common/graph/functional_node.py
@@ -1,5 +1,6 @@
-from typing import Dict, Any, Tuple, List
+from typing import Dict, Any, Tuple, Type
 
+from model_compression_toolkit.constants import FOUND_TF
 from model_compression_toolkit.core.common.graph.base_node import BaseNode
 import numpy as np
 
@@ -71,3 +72,17 @@ class FunctionalNode(BaseNode):
         :return: the node's functional_op
         """
         return self.functional_op
+
+    def is_match_type(self, _type: Type) -> bool:
+        """
+        Check if input type matches the node type, either in instance type or in type name. Checking the
+        name string is required because of function types changes that occurred in TF 2.15.
+
+        Args:
+            _type: other node type
+        Returns:
+            Whether _type matches the self node type
+
+        """
+        names_match = _type.__name__ == self.type.__name__ if FOUND_TF else False
+        return super().is_match_type(_type) or names_match
--- a/model_compression_toolkit/core/common/graph/functional_node.py
+++ b/model_compression_toolkit/core/common/graph/functional_node.py
@@ -76,7 +76,9 @@ class FunctionalNode(BaseNode):
     def is_match_type(self, _type: Type) -> bool:
         """
         Check if input type matches the node type, either in instance type or in type name. Checking the
-        name string is required because of function types changes that occurred in TF 2.15.
+        name string is required because of function types changes that occurred in TF 2.15, because it
+        changes the "function" attribute object (e.g. a different tf.add function that will fail the
+        equal operation).
 
         Args:
             _type: other node type
--- a/model_compression_toolkit/core/common/quantization/node_quantization_config.py
+++ b/model_compression_toolkit/core/common/quantization/node_quantization_config.py
@@ -410,7 +410,6 @@ class NodeWeightsQuantizationConfig(BaseNodeQuantizationConfig):
         for attr in node_attrs_list:
             if isinstance(attr, int):
                 # this is a positional attribute, so it needs to be handled separately.
-                # we assume that a positional attribute is quantized with the default configuration provided in the TPC.
                 self.pos_attributes_config_mapping[attr] = WeightsAttrQuantizationConfig(qc=qc,
                                                                                          weights_attr_cfg=op_cfg.default_weight_attr_config,
                                                                                          weights_channels_axis=weights_channels_axis)
--- a/model_compression_toolkit/core/common/quantization/node_quantization_config.py
+++ b/model_compression_toolkit/core/common/quantization/node_quantization_config.py
@@ -265,8 +265,6 @@ class WeightsAttrQuantizationConfig:
         self.enable_weights_quantization = weights_attr_cfg.enable_weights_quantization
         self.l_p_value = qc.l_p_value
 
-
-
     @property
     def weights_error_method(self) -> QuantizationErrorMethod:
         """
@@ -413,8 +411,6 @@ class NodeWeightsQuantizationConfig(BaseNodeQuantizationConfig):
             if isinstance(attr, int):
                 # this is a positional attribute, so it needs to be handled separately.
                 # we assume that a positional attribute is quantized with the default configuration provided in the TPC.
-                if op_cfg.default_weight_attr_config.enable_weights_quantization:
-                    Logger.critical(f"Quantizing constant weights is not supported.")
                 self.pos_attributes_config_mapping[attr] = WeightsAttrQuantizationConfig(qc=qc,
                                                                                          weights_attr_cfg=op_cfg.default_weight_attr_config,
                                                                                          weights_channels_axis=weights_channels_axis)
--- a/model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
+++ b/model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
+from typing import Dict
 import numpy as np
 from sklearn.cluster import KMeans
 
@@ -37,7 +38,7 @@ def lut_kmeans_tensor(tensor_data: np.ndarray,
                       n_iter: int = 10,
                       min_threshold: float = MIN_THRESHOLD,
                       quant_error_method: qc.QuantizationErrorMethod = None,
-                      is_symmetric=False) -> dict:
+                      is_symmetric: bool = False) -> Dict:
     """
     The quantizer first finds the closest max value per channel of tensor_data.
     Now, we divide tensor_data with the threshold vector per channel. In addition, we scale the result to the range
@@ -94,7 +95,7 @@ def lut_kmeans_histogram(bins: np.ndarray,
                          constrained: bool = True,
                          n_iter: int = 20,
                          min_threshold: float = MIN_THRESHOLD,
-                         quant_error_method: qc.QuantizationErrorMethod = qc.QuantizationErrorMethod.MSE) -> dict:
+                         quant_error_method: qc.QuantizationErrorMethod = qc.QuantizationErrorMethod.MSE) -> Dict:
     """
     Finds quantization cluster points for non-uniform activation quantization.
     The quantizer first finds the closest power-of-two number to the max value of the given histogram,
--- a/model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
+++ b/model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
@@ -19,7 +19,8 @@ from sklearn.cluster import KMeans
 
 import model_compression_toolkit.core.common.quantization.quantization_config as qc
 from model_compression_toolkit.constants import LUT_VALUES, MIN_THRESHOLD, SCALE_PER_CHANNEL, \
-    LUT_VALUES_BITWIDTH, THRESHOLD
+    LUT_VALUES_BITWIDTH, THRESHOLD, NUM_QPARAM_HESSIAN_SAMPLES
+from model_compression_toolkit.core.common.hessian import HessianInfoService
 from model_compression_toolkit.core.common.quantization.quantizers.quantizers_helpers import \
     max_power_of_two, int_quantization_with_threshold
 from model_compression_toolkit.core.common.quantization.quantization_params_generation.symmetric_selection import \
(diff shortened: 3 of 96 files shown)


model_compression_toolkit/exporter/model_wrapper/fw_agnostic/get_inferable_quantizers.py
```diff
@@ -32,18 +32,25 @@ def get_inferable_quantizers(node: BaseNode,
     Returns:
         weight_quantizers: A dictionary between a weight's name to its quantizer.
         activation_quantizers: A list of activations quantization, one for each layer output.
+        weight_values: A dictionary between a weight's name to its value. Relevant for positional weights only.
     """
 
     weight_quantizers = {}
     activation_quantizers = []
+    weight_values = None
```
ofirgo, line 40: Can you init this with an empty dict and then you can skip the `if weight_values is None:` part later and just update the dictionary with a key-value? would it be a problem to return an empty dict instead of None?

model_compression_toolkit/core/common/graph/functional_node.py
```diff
@@ -71,3 +72,17 @@ def type(self):
...
+        Check if input type matches the node type, either in instance type or in type name. Checking the
+        name string is required because of function types changes that occurred in TF 2.15.
+
+        Args:
+            _type: other node type
+        Returns:
+            Whether _type matches the self node type
+
+        """
+        names_match = _type.__name__ == self.type.__name__ if FOUND_TF else False
```
ofirgo, line 87: Very nice!
I just think that it needs more explaining (besides the one sentence mentioning tf 2.15 in the function's documentation)

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -413,8 +411,6 @@ def __init__(self, qc: QuantizationConfig,
             if isinstance(attr, int):
                 # this is a positional attribute, so it needs to be handled separately.
                 # we assume that a positional attribute is quantized with the default configuration provided in the TPC.
-                if op_cfg.default_weight_attr_config.enable_weights_quantization:
```
ofirgo, line 416: maybe change the log message to info and mention that this is an experimental feature

model_compression_toolkit/core/common/quantization/quantization_params_generation/lut_kmeans_params.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
+from typing import Dict
```
ofirgo, line 16: Just out of curiosity, how do you get to change the typehints on this unrelated file?
(In general, I think that it is preferred to not "fix" things that are unrelated to the PR because it overloads the changes and messes up the file's tracking, the change would appear with a commit message that is completely unrelated to the change that is made in the file)

model_compression_toolkit/core/common/similarity_analyzer.py
```diff
@@ -235,7 +235,7 @@ def compute_kl_divergence(float_tensor: np.ndarray, fxp_tensor: np.ndarray, batc
                           axis: int = None) -> float:
     """
     Compute the similarity between two tensor using KL-divergence.
-    The returned values is between 0 to 1: the smaller returned value,
+    The returned values is between 0 and 1: the smaller returned value,
```
ofirgo, line 238: same regarding unrelated changes

model_compression_toolkit/core/keras/reader/node_builder.py
```diff
@@ -67,12 +67,23 @@ def get_kwargs2index(tfoplambda_layer: TFOpLambda) -> Dict[str, int]:
     Returns:
         A dictionary with argument number and index: {arg_name: arg_index}.
     """
-    if tfoplambda_layer.function in [tf.add, tf.subtract, tf.divide, tf.truediv, tf.multiply, tf.pow,
-                                     tf.matmul, tf.image.crop_and_resize, tf.image.combined_non_max_suppression] or \
-            tfoplambda_layer.symbol in ['__operators__.add', 'math.add', 'math.multiply', 'linalg.matmul', 'concat']:
-        return {arg_name: i for i, arg_name in enumerate(tf_inspect.getfullargspec(tfoplambda_layer.function).args)}
-    else:
-        return {}
+    kwargs2index = {tf.add: {'x': 0, 'y': 1},
```
ofirgo, line 70: please add some explanation regarding this hardcoded mapping and why we need to work with it like this (I can see a scenario where in a few months we will look at this code and think to ourselves "why did we implement it with this constant dictionary instead of going over the function's argument list).
Also explain the if statements and the difference between the first and the second.
In addition, mention that in order to add support in other functional/positional ops/vars they will need to be added to this mapping.

ofirgo (commented):
Didn't go through everything, but some questions/notes in the meantime

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v2/tp_model.py
```diff
@@ -151,6 +152,14 @@ def generate_tp_model(default_config: OpQuantizationConfig,
     # this configuration will be used for the operation quantization:
     default_configuration_options = tp.QuantizationConfigOptions([default_config])
 
+    # const configuration is similar to the default eight bit configuration except for PoT
+    # quantization method for the constant.
```
ofirgo, line 156: I think we should elaborate more here, something like:
` # Create a QuantizationConfigOptions for quantizing constants in functional ops.
... (what you wrote)....
Since the constants are not named attributes of the layer, we use the default_weight_attr_config to define the desired quantization properties for them.
`

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v2/tp_model.py
```diff
@@ -151,6 +152,14 @@ def generate_tp_model(default_config: OpQuantizationConfig,
...
+    const_config = default_config.clone_and_edit(
+        default_weight_attr_config=default_config.default_weight_attr_config.clone_and_edit(
```
ofirgo, line 158: don't you think that the OpQuantizationConfig should be defined under get_op_quantization_configs like the other op configs and the the QuantizationConfigOptions should be defined here similarly to default_configuration_options ?

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v2/tp_model.py
```diff
@@ -32,7 +32,7 @@ def get_tp_model() -> TargetPlatformModel:
     NOTE: in order to generate a target platform model with different configurations but with the same Operators Sets
     (for tests, experiments, etc.), use this method implementation as a test-case, i.e., override the
     'get_op_quantization_configs' method and use its output to call 'generate_tp_model' with your configurations.
-    This version enables metadata by default
+    This version enables metadata by default.
```
ofirgo, line 35: I'm writing this comment here but it is general to this part: what about pytorch TPCs? why aren't they updated?

tests/keras_tests/feature_networks_tests/test_features_runner.py
```diff
@@ -537,6 +538,16 @@ def test_linear_collapsing(self):
         SixConv2DCollapsingTest(self).run_test()
         Op2DAddConstCollapsingTest(self).run_test()
 
+    def test_const_quantization(self):
```
ofirgo, line 541: maybe add tests for mixed precision and gptq for a model with const quantization?

In general, since we are not updating the default TPC to v2, I think that we need to think of a way to test more MCT features with const quantization enabled, otherwise we won't have a way to identify if something is broken due to the added const quantization (maybe change the tests TPCs the use "latest" to use "v2". or something else if you have a better idea)

tests/pytorch_tests/model_tests/feature_models/const_quantization_test.py
```diff
@@ -0,0 +1,88 @@
...
+        self.const = const
+        self.input_reverse_order = input_reverse_order
+
+    def generate_inputs(self):
+        return [np.random.random(in_shape)+1 for in_shape in self.get_input_shapes()]
+
+    def get_tpc(self):
+        return mct.get_target_platform_capabilities(PYTORCH, IMX500_TP_MODEL, "v2")
+
+    # def get_quantization_config(self):
```
ofirgo, line 63: remove comment

tests/keras_tests/feature_networks_tests/feature_networks/const_quantization_test.py
```diff
@@ -0,0 +1,82 @@
...
+                else:
+                    x = self.layer(x, self.const)
+        return tf.keras.models.Model(inputs=inputs, outputs=x)
+
+    def compare(self, quantized_model, float_model, input_x=None, quantization_info=None):
+        y = float_model.predict(input_x)
+        y_hat = quantized_model.predict(input_x)
+        self.unit_test.assertTrue(y.shape == y_hat.shape, msg=f'out shape is not as expected!')
+        cs = cosine_similarity(y, y_hat)
+        self.unit_test.assertTrue(np.isclose(cs, 1, atol=0.001), msg=f'fail cosine similarity check:{cs}')
```
ofirgo, line 77: please think of a different way to test this instead of using CS... it is very unstable (especially if you needed to increase the tolerance to 0.001)

tests/pytorch_tests/model_tests/feature_models/const_quantization_test.py
```diff
@@ -0,0 +1,88 @@
...
+            return ConstQuantizationNet(self.func, self.const)
+
+    def compare(self, quantized_model, float_model, input_x=None, quantization_info=None):
+        in_torch_tensor = to_torch_tensor(input_x[0])
+        set_model(float_model)
+        y = float_model(in_torch_tensor)
+        y_hat = quantized_model(in_torch_tensor)
+        self.unit_test.assertTrue(y.shape == y_hat.shape, msg=f'out shape is not as expected!')
+        cs = cosine_similarity(torch_tensor_to_numpy(y), torch_tensor_to_numpy(y_hat))
+        self.unit_test.assertTrue(np.isclose(cs, 1, atol=1e-1), msg=f'fail cosine similarity check: {cs}')
```
ofirgo, line 81: Same (CS). here it's even worse, if the tolerance needs to be 1e-1, are you sure that there isn't anything wrong with the results?
(by the way, I'm not sure how this test even works, given that you didn't add const quantization in the pytorch TPCs)

ofirgo (changes requested):

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v2/tp_model.py
```diff
@@ -151,6 +152,14 @@ def generate_tp_model(default_config: OpQuantizationConfig,
     # this configuration will be used for the operation quantization:
     default_configuration_options = tp.QuantizationConfigOptions([default_config])
 
+    # const configuration is similar to the default eight bit configuration except for PoT
+    # quantization method for the constant.
+    const_config = default_config.clone_and_edit(
+        default_weight_attr_config=default_config.default_weight_attr_config.clone_and_edit(
+            weights_quantization_method=tp.QuantizationMethod.POWER_OF_TWO,
+            enable_weights_quantization=True))
```
reuvenperetz, line 160: Is the quantization per-channel or per-tensor?

model_compression_toolkit/exporter/model_wrapper/fw_agnostic/get_inferable_quantizers.py
```diff
@@ -32,18 +32,25 @@ def get_inferable_quantizers(node: BaseNode,
     Returns:
         weight_quantizers: A dictionary between a weight's name to its quantizer.
         activation_quantizers: A list of activations quantization, one for each layer output.
+        weight_values: A dictionary between a weight's name to its value. Relevant for positional weights only.
```
reuvenperetz, line 35: Fix comment. Also, why not set the default value of attributes_name to None instead of a mutable list?

model_compression_toolkit/exporter/model_wrapper/fw_agnostic/get_inferable_quantizers.py
```diff
@@ -32,18 +32,25 @@ def get_inferable_quantizers(node: BaseNode,
...
 
     weight_quantizers = {}
     activation_quantizers = []
+    weight_values = None
 
     for attr in attributes_names:
         if node.is_weights_quantization_enabled(attr):
             weight_quantizer = get_weights_quantizer_for_node(node, attr)
             weight_quantizers[attr] = weight_quantizer
+            if isinstance(attr, int):  # for positional weights we need to extract the weight's value.
```
reuvenperetz, line 46: Is it possible to do it in a separate function? It seems very weird the combination of returned values and I don't see why they should be coupled.

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v2/tp_model.py
ofirgo: Following this comment:
- It should be per-tensor.
- you need to make sure that you are copying `eight_bits_default` config not `linear_eight_bits`.
- I think that the `weights_quantization_method` in `default_weight_attr_config`  is already defined to be POWER_OF_TWO, so you don't need to edit it (verify this before changing). Also the per-tensor property.

reuvenperetz (approved):

ofirgo (approved):

PR #1053, opened 2024-05-02
Increase coverage

reuvenperetz:
## Pull Request Description:
Improve coverage:

* Raise threshold test to 97%
* Add "no cover" in cases of "internal" errors, some "repr" methods, and imports that are dependent on the framework version.
* Remove parts that are unused: identity_wrapper, _generate_images_batch.
* Fix pytorch_device_config to return valid for "CPU" and add tests.
* Fix lut_fake_quant to use nbits and add test.
* Add tests to edge matcher, logger, torch_utils, activation quantization functions, immutable class, scheduler step keras optimizer, and metadata in onnx model.
* Use MP for activations in tensorboard test.
* Fix the name of the concat threshold test to have the prefix "test_".


tests/keras_tests/function_tests/test_activation_quantization_functions.py
```diff
@@ -0,0 +1,64 @@
...
+
+import numpy as np
+
+from model_compression_toolkit.constants import THRESHOLD, SIGNED, RANGE_MIN, RANGE_MAX
+
+from model_compression_toolkit.core.keras.quantizer.fake_quant_builder import symmetric_quantization, \
+    uniform_quantization
+
+
+class TestActivationQuantizationFunctions(unittest.TestCase):
```
ofirgo, line 26: How did we miss writing tests to this? 
Isn't it tested in test_quantization_configurations.py? (if not, maybe we should add it there instead of in this separate test)

tests/pytorch_tests/function_tests/test_activation_quantization_functions.py
```diff
@@ -0,0 +1,66 @@
...
+
+import torch
+
+from model_compression_toolkit.constants import THRESHOLD, SIGNED, RANGE_MIN, RANGE_MAX
+
+from model_compression_toolkit.core.pytorch.quantizer.fake_quant_builder import symmetric_quantization, \
+    uniform_quantization
+
+
+class TestActivationQuantizationFunctions(unittest.TestCase):
```
ofirgo, line 28: same as in keras

tests/pytorch_tests/function_tests/test_torch_utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+class TestTorchUtils(unittest.TestCase):
+    def setUp(self):
+        # Setup any common resources
+        self.numpy_array = np.array([1.0, 2.0, 3.0])
+        self.torch_tensor = torch.tensor([1.0, 2.0, 3.0])
+        self.list_of_numbers = [1, 2, 3]
+        self.tuple_of_numbers = (1, 2, 3)
+
+    @patch('model_compression_toolkit.core.pytorch.pytorch_device_config.get_working_device')
```
ofirgo, line 32: What is the purpose of this notation?

ofirgo (changes requested):

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -65,10 +65,7 @@ def __repr__(self) -> str:
         """
         Returns: String to display a NodeQuantizationConfig object.
         """
-        repr_str = ''
-        for k, v in self.__dict__.items():
-            repr_str += f'{k}: {v}\n'
-        return repr_str
+        return ''.join(f'{k}: {v}\n' for k, v in self.__dict__.items())  # pragma: no cover
```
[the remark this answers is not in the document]
reuvenperetz: This is mainly used during debugging. I can add a test that checks it, but I don't think this should be tested tbh.

model_compression_toolkit/core/common/quantization/quantization_config.py
```diff
@@ -124,7 +124,7 @@ def __init__(self,
         self.concat_threshold_update = concat_threshold_update
 
     def __repr__(self):
-        return str(self.__dict__)
+        return str(self.__dict__)  # pragma: no cover
```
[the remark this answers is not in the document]
reuvenperetz: Same. Because it's for debugging purposes.

model_compression_toolkit/core/common/substitutions/remove_identity.py
```diff
@@ -36,7 +37,8 @@ def remove_identity_node(graph: Graph,
 
     # Ensure there is exactly one predecessor; otherwise, do nothing.
     if len(prev_identity_nodes) != 1:
-        return graph
+        Logger.warning(f"Identity node {node} have {len(prev_identity_nodes)} inputs, while expected to have one. Skipping remove identity substitution.") # pragma: no cover
+        return graph  # pragma: no cover
```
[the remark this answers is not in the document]
reuvenperetz: I don't think you can create a real identity layer with multiple inputs. I can create some graph that demonstrates it, but it's not gonna be an actual real NN.

model_compression_toolkit/core/keras/back2framework/mixed_precision_model_builder.py
```diff
@@ -17,9 +17,9 @@
 from packaging import version
 import tensorflow as tf
 if version.parse(tf.__version__) >= version.parse("2.13"):
-    from keras.src.engine.base_layer import Layer
+    from keras.src.engine.base_layer import Layer  # pragma: no cover
```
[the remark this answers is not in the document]
reuvenperetz: The lines we cover depend on the version. We cover the lines that are used. So I'll just remove the "no cover" from the lines we cover (the "if" block).

tests/pytorch_tests/function_tests/test_torch_utils.py
reuvenperetz: The @patch is used to temporarily replace the behavior of a function or object during the execution of a unit test. In this case, when the test encounters a call get_working_device, it won't execute the actual function, but the mock object.

tests/keras_tests/function_tests/test_activation_quantization_functions.py
reuvenperetz: You're correct. test_quantization_configurations checks it, but it's only for Keras (and I added tests for Pytorch). So I can either create a similar test_quantization_configurations for pytorch, or these kind of tests for Keras. I think the first, is a better idea actually.

reuvenperetz: They are not tested in test_quantization_configurations. These functions are used in some cases like MP for activations and SNC (which are all tested only with PoT quantizer, currently).

ofirgo (approved):

PR #1073, opened 2024-05-16
Small refactor to PyTorch model reader to handle tensors received in node's kwargs.

elad-c:
## Pull Request Description:
This PR fixes a bug that the "Const Representation" PR reveiled, that input tensors to a node through it's kwargs were not handled correctly.

### DIFF
--- a/model_compression_toolkit/core/pytorch/reader/graph_builders.py
+++ b/model_compression_toolkit/core/pytorch/reader/graph_builders.py
@@ -53,6 +53,25 @@ def extract_holder_weights(constant_name, node_target, model, weights, to_numpy)
     return weights
 
 
+def get_node_args_spec(node: Node) -> List[str]:
+    """
+    Return a list of the node's function args & kwargs, so we can insert the kwarg as arg
+    Args:
+        node: Torch.fx.Node to extract target from, which the function the Node executes.
+
+    Returns: A list of args and kwargs names.
+
+    """
+    if node.target in [torch.nn.functional.conv1d, torch.nn.functional.conv2d, torch.nn.functional.conv3d]:
+        return ['input', 'weight', 'bias', 'stride', 'padding', 'dilation', 'groups']
+    elif node.target in [torch.nn.functional.conv_transpose1d, torch.nn.functional.conv_transpose2d,
+                         torch.nn.functional.conv_transpose3d]:
+        return ['input', 'weight', 'bias', 'stride', 'padding', 'output_padding', 'groups', 'dilation']
+    elif node.target in [torch.nn.functional.linear]:
+        return ['input', 'weight', 'bias']
+    return inspect.getfullargspec(node.target).args
+
+
 def nodes_builder(model: GraphModule,
                   module_dict: Dict,
                   to_numpy: Callable) -> Tuple[List, List, List, Dict]:
@@ -163,10 +182,15 @@ def nodes_builder(model: GraphModule,
                 framework_attr_filtered[k] = v
         framework_attr = framework_attr_filtered
 
-        # filter Nodes from node kwargs, we replace these attributes with nx graph nodes
+        # filter Nodes from node kwargs, we replace these attributes with nx graph nodes. The Node kwargs
+        # are saved in _args_from_kwargs so the index of the kwargs can be inserted to the call args so
+        # the tensor indices can be extracted.
         node_kwargs = {}
+        _args_from_kwargs = {}
         for k, v in node.kwargs.items():
-            if not isinstance(v, torch.fx.node.Node):
+            if isinstance(v, torch.fx.node.Node):
+                _args_from_kwargs[k] = v
+            else:
                 node_kwargs[k] = v
 
         # initiate graph nodes
@@ -178,11 +202,22 @@ def nodes_builder(model: GraphModule,
                                                  node.args[0].op == PLACEHOLDER and node.args[0].meta[TYPE] in (list, tuple))
             tensor_input_index = []
             op_call_args = list(node.args)
+            for arg_name, arg_val in _args_from_kwargs.items():
+                _args = get_node_args_spec(node)
+                if arg_name in _args:
+                    arg_index = _args.index(arg_name)
+                    # TODO: This code verifies the kwargs with Node values that are used as args are in the correct order.
+                    # (i.e. check that the kwarg with Node is the first of the kwargs)
+                    if len(op_call_args) != arg_index:
+                        Logger.warning("A kwarg with value of Node should be the first after the arguments. "
+                                       "This might fail the model builder later...")
+                    op_call_args.insert(arg_index, arg_val)
+
             if inputs_as_list:
                 op_call_args.pop(0)
             else:
                 for in_node in node.all_input_nodes:
-                    for i, arg in enumerate(node.args):
+                    for i, arg in enumerate(op_call_args):
                         if arg == in_node:
                             tensor_input_index.append(i)
 
--- a/model_compression_toolkit/core/pytorch/reader/graph_builders.py
+++ b/model_compression_toolkit/core/pytorch/reader/graph_builders.py
@@ -55,9 +55,12 @@ def extract_holder_weights(constant_name, node_target, model, weights, to_numpy)
 
 def get_node_args_spec(node: Node) -> List[str]:
     """
-    Return a list of the node's function args & kwargs, so we can insert the kwarg as arg
+    This function builds a list of all layer operation arguments. It flattens and combines both the args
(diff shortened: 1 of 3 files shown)


model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -53,6 +53,25 @@ def extract_holder_weights(constant_name, node_target, model, weights, to_numpy)
...
+def get_node_args_spec(node: Node) -> List[str]:
+    """
+    Return a list of the node's function args & kwargs, so we can insert the kwarg as arg
+    Args:
+        node: Torch.fx.Node to extract target from, which the function the Node executes.
+
+    Returns: A list of args and kwargs names.
+
+    """
+    if node.target in [torch.nn.functional.conv1d, torch.nn.functional.conv2d, torch.nn.functional.conv3d]:
```
ofirgo, line 65: why the special treatment for these specific layer types?

model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -53,6 +53,25 @@ def extract_holder_weights(constant_name, node_target, model, weights, to_numpy)
     return weights
 
 
+def get_node_args_spec(node: Node) -> List[str]:
+    """
+    Return a list of the node's function args & kwargs, so we can insert the kwarg as arg
+    Args:
+        node: Torch.fx.Node to extract target from, which the function the Node executes.
```
ofirgo, line 60: what does "which the function the Node executes" mean?

model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -53,6 +53,25 @@ def extract_holder_weights(constant_name, node_target, model, weights, to_numpy)
     return weights
 
 
+def get_node_args_spec(node: Node) -> List[str]:
+    """
+    Return a list of the node's function args & kwargs, so we can insert the kwarg as arg
```
ofirgo, line 58: give some more explanation, maybe give an example of a problematic case.
the terms "function args", and "insert the kwargs as args" are not clear, I know what you mean just because you explained it to me personally.
So elaborate a little more, something like:
"This function builds an ordered (?) list of layer operation arguments. It flattens and combines both the args list and the kwargs dictionary for a single list, specifically, enables to treat the kwargs as arguments. This is important for the case where functional input (or find a better term), which is a dynamic output of another layer's operation (again, fix my explanation if it is not correct), is passed to the layer as a key argument".

Also, "return**s**", and a '.' at the end of a sentence.

model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -163,10 +182,15 @@ def nodes_builder(model: GraphModule,
                 framework_attr_filtered[k] = v
         framework_attr = framework_attr_filtered
 
-        # filter Nodes from node kwargs, we replace these attributes with nx graph nodes
+        # filter Nodes from node kwargs, we replace these attributes with nx graph nodes. The Node kwargs
```
ofirgo, line 185: maybe try to clarify this as well, I'm afraid it is written in a way that only you can understand because you have the explanation in your head.
I think that there is no escape from providing an example.

model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -178,11 +202,22 @@ def nodes_builder(model: GraphModule,
                                                  node.args[0].op == PLACEHOLDER and node.args[0].meta[TYPE] in (list, tuple))
             tensor_input_index = []
             op_call_args = list(node.args)
+            for arg_name, arg_val in _args_from_kwargs.items():
+                _args = get_node_args_spec(node)
+                if arg_name in _args:
+                    arg_index = _args.index(arg_name)
+                    # TODO: This code verifies the kwargs with Node values that are used as args are in the correct order.
```
ofirgo, line 209: what's the "todo"? even if it just so it will be clear what's going on here, add a sentence about how it should be eventually handled

model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -178,11 +202,22 @@ def nodes_builder(model: GraphModule,
...
+                    # (i.e. check that the kwarg with Node is the first of the kwargs)
+                    if len(op_call_args) != arg_index:
+                        Logger.warning("A kwarg with value of Node should be the first after the arguments. "
```
ofirgo, line 212: what is "value of Node"? (I think I know what it means because you should it to me in the debugger, but it is not clear)

model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -178,11 +202,22 @@ def nodes_builder(model: GraphModule,
...
+                                       "This might fail the model builder later...")
```
ofirgo, line 213: This might cause issues when trying to rebuild the model from the graph"

tests/pytorch_tests/model_tests/feature_models/const_representation_test.py
```diff
@@ -101,3 +101,39 @@ def __init__(self, unit_test):
 
     def create_networks(self):
         return ConstRepresentationMultiInputNet()
+
+
+class ConstRepresentationCodeNet(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv2d = nn.Conv2d(16, 16, 3, 2, padding=1)
+        # self.bn = torch.nn.BatchNorm2d(16)
```
ofirgo, line 110: remove?

tests/pytorch_tests/model_tests/feature_models/const_representation_test.py
```diff
@@ -101,3 +101,39 @@ def __init__(self, unit_test):
...
+        # TODO: Enable this code when PyTorch model reader is fixed for this case. Also, move this to model reader tests.
+        # y = torch.nn.functional.batch_norm(y[:, :, 0, :], self.bn.running_mean, self.bn.running_var,
+        #                                    momentum=0.2, eps=1e-6, bias=self.bn.bias)
+        # y = torch.unsqueeze(y, 2)
+        return torch.add(x, y)
+
+
+class ConstRepresentationCodeTest(ConstRepresentationTest):
+    """
+    This test checks the code representation code in the model reader and builder, not specifically
```
ofirgo, line 126: first "code" --> "const"

tests/pytorch_tests/model_tests/feature_models/const_representation_test.py
```diff
@@ -101,3 +101,39 @@ def __init__(self, unit_test):
...
+    with constants in the graph.
```
ofirgo, line 127: so explain what it checks (arguments order and parsing?)

tests/pytorch_tests/model_tests/test_feature_models_runner.py
```diff
@@ -252,6 +252,8 @@ def test_const_representation(self):
 
         ConstRepresentationMultiInputTest(self).run_test()
 
+        ConstRepresentationCodeTest(self, None, None).run_test()
```
ofirgo, line 255: what are the 2 Nones?

tests/pytorch_tests/model_tests/feature_models/const_representation_test.py
```diff
@@ -101,3 +101,39 @@ def __init__(self, unit_test):
 
     def create_networks(self):
         return ConstRepresentationMultiInputNet()
+
+
+class ConstRepresentationCodeNet(nn.Module):
```
ofirgo, line 106: maybe duplicate this test to check different arguments combinations? (namely, args and kwargs, dynamic, etc.)
do you think it would be valuable to test?

ofirgo (changes requested):
mainly documentation comments, some testing comments

model_compression_toolkit/core/pytorch/reader/graph_builders.py
elad-c: Because `inspect.getfullargspec(node.target)` fails for them.

tests/pytorch_tests/model_tests/feature_models/const_representation_test.py
elad-c: Added functional `Conv2d`, `ConvTranspose2d` & `Linear` to cover all options of `get_node_args_spec`

model_compression_toolkit/core/pytorch/reader/graph_builders.py
elad-c: only used here, so no need to add code for it.

PR #1082, opened 2024-05-21
Use normalize MSE in mixed precision sensitivity evaluation

eladc-git:
## Pull Request Description:
Use normalized MSE distance function in mixed precision sensitivity evaluation. For hessian-based sensitivity, simple MSE (not normalized) is used.

### DIFF
--- a/model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation.py
+++ b/model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation.py
@@ -89,12 +89,13 @@ class SensitivityEvaluation:
                                                       fw_impl.count_node_for_mixed_precision_interest_points,
                                                       quant_config.num_interest_points_factor)
 
-        self.ips_distance_fns, self.ips_axis = self._init_metric_points_lists(self.interest_points,
-                                                                              self.quant_config.use_hessian_based_scores is False)
+        # We use normalized MSE when not running hessian-based. For Hessian-based normalized MSE is not needed
+        # beacause hessian weights already do normalization.
+        use_normalized_mse = self.quant_config.use_hessian_based_scores is False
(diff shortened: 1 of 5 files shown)


model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation.py
```diff
@@ -89,10 +89,12 @@ def __init__(self,
                                                       fw_impl.count_node_for_mixed_precision_interest_points,
                                                       quant_config.num_interest_points_factor)
 
-        self.ips_distance_fns, self.ips_axis = self._init_metric_points_lists(self.interest_points)
+        self.ips_distance_fns, self.ips_axis = self._init_metric_points_lists(self.interest_points,
```
ofirgo, line 92: add a comment here explaining the default behavior (that we run normalized MSE when not running hessian-based, you can also explain why in a sentence), otherwise we won't be able to figure out why we did it if we'll look at this one year from now

tests/keras_tests/function_tests/test_sensitivity_metric_interest_points.py
```diff
@@ -145,7 +145,7 @@ def test_softmax_interest_point(self):
             if axis is None:
                 axis = sn.op_call_kwargs.get(AXIS)
 
-            distance_fn = KerasImplementation().get_node_distance_fn(layer_class=sn.layer_class,
+            distance_fn = KerasImplementation().get_mp_node_distance_fn(layer_class=sn.layer_class,
```
ofirgo, line 148: this change makes the lines not aligned

ofirgo (approved):
See minor comments.
In addition:
- Update @elad-c directly regarding adding this to the release notes because he already edited them for the release.
- Did you make sure that both cases are covered in our MP tests? if not, maybe we should add a dedicated test

PR #1086, opened 2024-05-26
reshape error fix

samuel-wj-chapman:
## Pull Request Description:
Solution works for all the problem models. However, a cleaner fix might be preferred.

Copied from below - error:
reshape() takes 2 positional arguments but 5 were given

for the following models
inception_next_small.sail_in1k
inception_next_tiny.sail_in1k
inception_next_base.sail_in1k
inception_next_base.sail_in1k_384

error occurs due to line 161 in [this inception model](https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/inception_next.py)

x = x.mul(self.gamma.reshape(1, -1, 1, 1))

arguments fed through incorrect format. change specifically looks for his case and resolves.


haihabi:
@ofirgo please take a look...

ofirgo:
Hi @samuel-wj-chapman , 
Can you please explain the error and when it occurs?

@elad-c can you please take a look, I think it's related to positional attributes

samuel-wj-chapman:
error
reshape() takes 2 positional arguments but 5 were given

for the following models
inception_next_small.sail_in1k
inception_next_tiny.sail_in1k
inception_next_base.sail_in1k
inception_next_base.sail_in1k_384

error occurs due to line 161 in [this inception model](https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/inception_next.py)

x = x.mul(self.gamma.reshape(1, -1, 1, 1))

arguments fed through incorrect format. change specifically looks for his case and resolves.

elad-c:
I think it's the same issue we has with `permute`. We handled it with a [substitution](model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/permute_call_method.py). Maybe we should also handle `reshape` in that substitution.
Anyway, please add a test for this case.

haihabi:
@elad-c and @samuel-wj-chapman  If we handle this kind of issue via [substitution](https://github.com/sony/model_optimization/pull/model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/permute_call_method.py)., then I suggest that here we take the same approach as well.

model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/reshape_call_method.py
```diff
@@ -0,0 +1,52 @@
...
+from model_compression_toolkit.core.common.graph.graph_matchers import NodeOperationMatcher
+from model_compression_toolkit.core import common
+from model_compression_toolkit.core.common.graph.base_graph import Graph
+from model_compression_toolkit.core.common.graph.base_node import BaseNode
+
+
+class ReshapeCallMethod(common.BaseSubstitution):
+    """
+    Find "permute" node to substitute new dimension argument if needed
+    """
```
Idan-BenAmi, line 26: typo "permute"

model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/reshape_call_method.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self):
+        """
+        Matches: 'permute' node
+        """
```
Idan-BenAmi, line 31: same in here: permute

elad-c (changes requested):
Can you "reuse" the existing substitution for `permute` and apply it for `reshape` too? Seems to be identical to the permute substitution except for the `NodeMatcher`.
Something like:
```
nodes = NodeOperationMatcher(permute) | NodeOperationMatcher(reshape)
super().__init__(matcher_instance=nodes)
```
would do the trick.

samuel-wj-chapman:
the pr failed test when i duplicated. the addition line  "if len(node.op_call_args)>1:"  was needed to prevent other correct instances of reshape from being substituted. @elad-c

elad-c:
Does it fail in the case of permute?

samuel-wj-chapman:
no :)

tests/pytorch_tests/model_tests/feature_models/reshape_substitution_test.py
elad-c, line 1: Maybe this file can be avoided. It's very similar to `permute_substitution_test.py`. Can we change it to `transform_fn_call_substitution_test.py` and test both `reshape` and `permute`?

Idan-BenAmi: Guessing it's also good to keep the 2 tests.. so we know we check both cases.. but not too critical either way.

samuel-wj-chapman: I feel maybe best to keep them separate as we needed to make a change to the original file to allow it to work with both transform types? eg - if isinstance(node.op_call_args[0], (list, tuple)).

elad-c (approved):

PR #1109, opened 2024-06-17
Verify bias is not None in GPTQ bias training

ofirgo:
## Pull Request Description:
GPTQ with bias train is failing in the rare case where a linear layer doesn't have a bias


reuvenperetz (approved):
Approved. But maybe you should add a unit test for this case (per framework)?

PR #1143, opened 2024-08-01
Update imx500 notebooks mobilenetv2 - ImageNet > Imagenette

alexander-sony:
## Pull Request Description:
updates to example_keras_mobilenetv2_for_imx500.ipynb
- pin versions of tf and mct
- use imagenette, a subset of imagenet - we exempilfy so lets keep dataset small. moreover, downloading and loading imagenette is so much more convinient since its part of torchvision.
- introduced imagenette preprocessor - needs labels remapping
- removed dependency on prepare_imagenet.sh
- moved some dataset constants to the top of tutorial
- added some asserts


Idan-BenAmi:
Hi @alexander-sony ,
Looks like a good approach. Some general comments:
1. Let's keep it as a new and separated tutorial for "Imagenette" dataset. (we still want the Imagenet tutorial)
2. In the Keras tutorial, we've tried to remove the requirement of torchvision (to ease the environment setup). At first glance, it looks like the Imagenette dataset can be downloaded as it is (only requires extraction), without any further preparation. Therefore you may be able to easily remove the torchvision requirement.
3. we don't want to fix the MCT version - since we test those tutorials on the nightly version.
4. we suggest to define other versions like TF_VER in the tutorials/notebooks/imx500_notebooks/__init__.py, where all future tutorial can import the same version.

Thanks
Idan

Idan-BenAmi (commented):
Hi @alexander-sony, Please see my comments on the conversation. The main thing is to add your tutorial as an additional notebook and not instead of the current tutorial.

alexander-sony:
TBD

PR #1154, opened 2024-08-07
Check for TF version smaller than 2.16

lapid92:
## Pull Request Description:
Verify if the installed TensorFlow versions are below 2.16. If the user attempts to use MCT with an unsupported TensorFlow version, they will receive an informative message.


model_compression_toolkit/verify_packages.py
```diff
@@ -0,0 +1,18 @@
+import importlib
```
lior-dikstein, line 1: Add licence

lior-dikstein (approved):

PR #1155, opened 2024-08-07
fix find duplicates in qconfig

eladc-git:
## Pull Request Description:
Use names in is_match method only in tensorflow operations.

### DIFF
--- a/model_compression_toolkit/core/common/graph/functional_node.py
+++ b/model_compression_toolkit/core/common/graph/functional_node.py
@@ -1,6 +1,5 @@
 from typing import Dict, Any, Tuple, Type, List, Union
 
-from model_compression_toolkit.constants import FOUND_TF
 from model_compression_toolkit.core.common.graph.base_node import BaseNode
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -553,7 +553,7 @@ def get_qco(self, tpc: TargetPlatformCapabilities) -> QuantizationConfigOptions:
         # Extract qco with is_match_type to overcome mismatch of function types in TF 2.15
         matching_qcos = [_qco for _type, _qco in tpc.layer2qco.items() if self.is_match_type(_type)]
         if matching_qcos:
-            if len(matching_qcos) > 1:
+            if len(matching_qcos) > 1 and matching_qcos[0] != tpc.tp_model.default_qco:
```
elad-c, line 556: 1. Must the deafult be at index 0?
2. Add a test to simulate it.

elad-c (changes requested):

model_compression_toolkit/core/common/graph/base_node.py
ofirgo, line 556: I don't think that this solves the issue correctly.
I suspect that the case generating this issue is when there is a pytorch function that has two aliases, and that one of them appears in a TPC operation set and the other is not, thus returning the default qco.

Please try to reproduce the issue, identify the pytorch layer that has duplicate configs and let's think of a more robust solution.

elad-c:
Issue will be fixed in another PR.

PR #1204, opened 2024-09-09
GraphBuilder - verify node name is not reserved method of nn.module

Idan-BenAmi:


Idan-BenAmi:
This PR intends to rename reserved name of node on the Graph such as "to" in order to avoid confusion between node of the output model and the PyTorch operation "to".
We still consider if the operation "to" as part of input model is valid MCT input, therefore, setting this PR to draft.

PR #1213, opened 2024-09-11
Fix PyTorch model reader

elad-c:
## Pull Request Description:
Fix PyTorch model reader for nodes with list input (e.g. `concat`), when both inputs are the same tensor (i.e. `concat([x, x], dim=2)`).

### DIFF
--- a/model_compression_toolkit/core/pytorch/reader/graph_builders.py
+++ b/model_compression_toolkit/core/pytorch/reader/graph_builders.py
@@ -80,6 +80,8 @@ def _build_input_alloc_and_call_args(n: Node, input_tensors_in_node_kwargs: Dict
     tensor_input_alloc = []
     op_call_args = list(n.args)
     if inputs_as_list:
+        # input tensors are a list in the first argument -> remove from op_call_args and go over
+        # the tensors in that list.
         _args = op_call_args.pop(0)
     else:
         _args = n.args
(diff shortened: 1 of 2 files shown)


model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -80,16 +80,17 @@ def _build_input_alloc_and_call_args(n: Node, input_tensors_in_node_kwargs: Dict
     tensor_input_alloc = []
     op_call_args = list(n.args)
     if inputs_as_list:
-        op_call_args.pop(0)
+        _args = op_call_args.pop(0)
```
ofirgo, line 83: can you add a comment explaining the two cases? why in one case you take from op_call_args and the other case you take from n.args?

tests/pytorch_tests/model_tests/feature_models/activation_16bit_test.py
```diff
@@ -60,6 +60,12 @@ def create_networks(self):
         return Activation16BitNet()
 
     def compare(self, quantized_model, float_model, input_x=None, quantization_info=None):
+        x = torch.from_numpy(input_x[0].astype('float32'))
+        out_f = float_model(x)
+        quantized_model = quantized_model.to('cpu')
```
ofirgo, line 65: don't we have constant for this ('cpu')?

ofirgo (approved):
Jsut two minor comments

PR #1242, opened 2024-10-14
Refactor hessian service and torch sample layer attention. 

irenaby:
## Pull Request Description:
Refactor hessian service.
Add data util for torch.
Update GPTQ trainer - use new data loader for samples and corresponding loss weights, rewrite SLA. 

### DIFF
--- a/model_compression_toolkit/core/common/hessian/hessian_info_service.py
+++ b/model_compression_toolkit/core/common/hessian/hessian_info_service.py
@@ -43,7 +43,7 @@ class HessianCache:
     def __init__(self):
         self._data: Dict[Query, Tensor] = {}
 
-    def update(self, layers_hessians, request) -> int:
+    def update(self, layers_hessians: Dict[str, np.ndarray], request: HessianScoresRequest) -> int:
         """
         Updates the cache with new hessians estimations.
 
--- a/model_compression_toolkit/core/common/hessian/hessian_info_service.py
+++ b/model_compression_toolkit/core/common/hessian/hessian_info_service.py
@@ -23,7 +23,6 @@ from model_compression_toolkit.core.common.hessian.hessian_scores_request import
 
 
 # type hints aliases
-SampleHash = str
 LayerName = str
 Tensor = Any
 
@@ -37,6 +36,7 @@ class Query:
 
 
 class HessianCache:
+    """ Hessian cache """
     def __init__(self):
         self._data: Dict[Query, Tensor] = {}
 
--- a/model_compression_toolkit/core/common/hessian/hessian_info_service.py
+++ b/model_compression_toolkit/core/common/hessian/hessian_info_service.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 from dataclasses import dataclass
-from typing import List, Dict, Any, Tuple
+from typing import List, Dict, Tuple, TYPE_CHECKING
 
 import numpy as np
 
@@ -21,10 +21,13 @@ from model_compression_toolkit.constants import HESSIAN_NUM_ITERATIONS
 from model_compression_toolkit.core.common.hessian.hessian_scores_request import HessianScoresRequest, HessianMode, \
     HessianScoresGranularity
 
+if TYPE_CHECKING:    # pragma: no cover
+    from model_compression_toolkit.core.common import BaseNode
+
 
 # type hints aliases
 LayerName = str
-Tensor = Any
+Tensor = np.ndarray
 
 
 @dataclass(eq=True, frozen=True)
@@ -63,7 +66,7 @@ class HessianCache:
 
         return min(n_nodes_samples)
 
-    def fetch_hessians(self, request: HessianScoresRequest) -> Tuple[Dict[LayerName, Tensor], Dict[LayerName, int]]:
+    def fetch_hessian(self, request: HessianScoresRequest) -> Tuple[Dict[LayerName, Tensor], Dict[LayerName, int]]:
         """
         Fetch available hessians per request and identify missing samples.
 
@@ -133,6 +136,9 @@ class HessianInfoService:
                       force_compute: bool = False) -> Dict[LayerName, Tensor]:
         """
         Fetch hessians per request.
+        If 'force_compute' is False, will first try to retrieve previously cached hessians. If no or not enough
+        hessians are found in the cache, will compute the remaining number of hessians to fulfill the request.
+        If 'force_compute' is True, will compute the hessians (use when you need hessians for specific inputs).
 
         Args:
             request: request per which to fetch the hessians.
@@ -140,7 +146,7 @@ class HessianInfoService:
                            If False, will look for cached hessians first.
 
         Returns:
-            A dictionary or a list of layers' hessian tensors of shape (samples, ...). The exact shape depends on the
+            A dictionary of layers' hessian tensors of shape (samples, ...). The exact shape depends on the
             requested granularity.
         """
         if request.n_samples is None and not force_compute:
@@ -166,7 +172,7 @@ class HessianInfoService:
         """ Purge the cached hessians. """
         self.cache.clear()
 
-    def _fetch_hessians_with_compute(self, request: HessianScoresRequest, n_iterations: int):
+    def _fetch_hessians_with_compute(self, request: HessianScoresRequest, n_iterations: int) -> Dict[LayerName, Tensor]:
         """
         Fetch pre-computed hessians for the request if available. Otherwise, compute the missing hessians.
 
@@ -177,13 +183,14 @@ class HessianInfoService:
         Returns:
             A dictionary from layers (by name) to their hessians.
         """
-        res, missing = self.cache.fetch_hessians(request)
+        res, missing = self.cache.fetch_hessian(request)
         if not missing:
             return res
 
         if request.data_loader is None:
-            raise ValueError('Not enough hessians are cached to fulfill the request, but data loader was not passed '
-                             'for additional computation.')
+            raise ValueError(f'Not enough hessians are cached to fulfill the request, but data loader was not passed '
+                             f'for additional computation. Requested {request.n_samples}, '
+                             f'available {min(missing.values())}.')
 
         orig_request = request
         # if some hessians were found generate a new request only for missing nodes.
@@ -191,11 +198,12 @@ class HessianInfoService:
             target_nodes = [n for n in orig_request.target_nodes if n.name in missing]
             request = request.clone(target_nodes=target_nodes)
         self._compute_hessians(request, n_iterations, count_by_cache=True)
-        res, missing = self.cache.fetch_hessians(request)
+        res, missing = self.cache.fetch_hessian(request)
         assert not missing
         return res
 
-    def _compute_hessians(self, request: HessianScoresRequest, n_iterations: int, count_by_cache: bool):
+    def _compute_hessians(self, request: HessianScoresRequest,
+                          n_iterations: int, count_by_cache: bool) -> Dict[LayerName, Tensor]:
         """
         Computes hessian estimation per request.
 
@@ -235,7 +243,8 @@ class HessianInfoService:
 
         if request.n_samples:
             if n_samples < request.n_samples:
-                raise ValueError('Not enough samples in the provided representative dataset')
+                raise ValueError(f'Could not compute the requested number of Hessians ({request.n_samples}), '
+                                 f'not enough samples in the provided representative dataset.')
 
             if n_samples > request.n_samples:
                 hess_per_layer = {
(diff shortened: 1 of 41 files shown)


model_compression_toolkit/core/pytorch/data_util.py
```diff
@@ -0,0 +1,161 @@
+# Copyright 2024 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo, line 1: If these function are only used as utils for the Hessian service then I think it's better to put it under "hessian" folder, to reduce the chance of using them for other purposes without paying attention

model_compression_toolkit/core/pytorch/data_util.py
```diff
@@ -0,0 +1,161 @@
...
+            data_gen_fn: data generator factory.
+            n_samples: target size of the dataset. If None, use all samples yielded by the data generator in one pass.
+        """
+        test_batch = next(data_gen_fn())
+        if not isinstance(test_batch, list):
+            raise TypeError(f'Data generator is expected to yield a list of tensors, got {type(test_batch)}')
+        self.orig_batch_size = test_batch[0].shape[0]
+
+        samples = []
+        for batch in data_gen_fn():
```
ofirgo, line 93: I can't locate this in the changes so I'll just ask - is this, or any of the other dataset generators defined here, run shuffle on the dataset? meaning, are the batches returned with different samples in each epoch or with the same samples in each batch?

Idealy, for GPTQ, we would want shuffling in the dataset

model_compression_toolkit/gptq/common/gptq_training.py
```diff
@@ -261,11 +207,9 @@ def build_gptq_model(self):
...
     @abstractmethod
-    def train(self, representative_data_gen: Callable):
+    def train(self):
         """
-        Train the quantized model using GPTQ training process
-        Args:
-            representative_data_gen: Dataset to use for inputs of the models.
+        Train the quantized model using GPTQ training process.
         """
         raise NotImplemented(f'{self.__class__.__name__} have to implement the '
```
ofirgo, line 214: NotImplementedError? (;

model_compression_toolkit/gptq/pytorch/gptq_training.py
```diff
@@ -70,6 +73,7 @@ def __init__(self,
             hessian_info_service: HessianInfoService to fetch info based on the hessian approximation of the float model.
         """
         def _get_total_grad_steps():
+            # TODO get it from the dataset
```
ofirgo, line 76: It's not possible after this refactor? (probably not but I'm just making sure)

model_compression_toolkit/gptq/pytorch/gptq_training.py
```diff
@@ -106,20 +111,84 @@ def _get_total_grad_steps():
...
+        hessians_tensor = torch.stack([layers_hessians[layer.name] for layer in self.compare_points], dim=1)    # samples X layers
+        assert hessians_tensor.shape[1] == len(self.compare_points)
+        loss_weights = list(hessians_tensor)
+        sla_train_dataset = FixedSampleInfoDataset(fixed_dataset.samples, loss_weights)
+
+        reg_weights = hessians_tensor.mean(dim=0)
+        # use collate to add a single value to each batch
+        collate_fn = get_collate_fn_with_extra_outputs(reg_weights)
+
+        return DataLoader(sla_train_dataset, batch_size=orig_batch_size, shuffle=True, collate_fn=collate_fn)
```
ofirgo, line 161: I see now the "shuffle" flag so it probably answers my previous question regarding the shuffle, but I'm keeping it open just to make sure

ofirgo (commented):
Still need to go over the tests but in the meanwhile leaving some comments

model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation.py
```diff
@@ -238,22 +238,20 @@ def _compute_hessian_based_scores(self) -> np.ndarray:
         """
         # Create a request for Hessian approximation scores with specific configurations
         # (here we use per-tensor approximation of the Hessian's trace w.r.t the node's activations)
+        fw_dataloader = self.fw_impl.convert_data_gen_to_dataloader(self.representative_data_gen,
```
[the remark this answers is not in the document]
irenaby: Not necessarily. The request should contain the dataloader for hessian computation with the appropriate batch size and that's something that the caller must supply. It definitely shouldn't be a responsibility of the request to convert python data generators into dataloader. BTW dataloader is just defined as iterable, so if the batch size was appropriate data generator instance could be passed as is. Neither request not the hessian service care what exactly dataloader is.

model_compression_toolkit/core/common/pruning/importance_metrics/lfh_importance_metric.py
```diff
@@ -120,22 +120,21 @@ def _get_entry_node_to_score(self, entry_nodes: List[BaseNode]) -> Dict[BaseNode
...
 
         # Initialize HessianInfoService for score computation.
+
         hessian_info_service = HessianInfoService(graph=self.float_graph,
-                                                  representative_dataset_gen=self.representative_data_gen,
                                                   fw_impl=self.fw_impl)
 
         # Fetch and process Hessian scores for output channels of entry nodes.
-        nodes_scores = []
-        for node in entry_nodes:
```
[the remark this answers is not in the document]
irenaby: Yes, I though we talked about this, but maybe only I thought that :) I can restore it. Gradient-wise it should be identical but there is a small difference that now the same random vector is used for all nodes whereas previously random vector was generated for each node.

model_compression_toolkit/core/common/quantization/quantization_params_generation/error_functions.py
```diff
@@ -377,23 +377,26 @@ def _get_sliced_histogram(bins: np.ndarray,
...
+                              num_hessian_samples: int,
+                              dataloader: Optional[Iterable]) -> List[List[np.ndarray]]:
     """
     Compute and retrieve Hessian-based scores for using during HMSE error computation.
 
     Args:
         node: The node to compute Hessian-based scores for.
         hessian_info_service: HessianInfoService object for retrieving Hessian-based scores.
         num_hessian_samples: Number of samples to approximate Hessian-based scores on.
+        dataloader: Data loader for computing Hessian-based scores. Can be None if hessians are expected to be available.
```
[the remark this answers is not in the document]
irenaby: It means they were already precomputed, I'll try to make it clearer

model_compression_toolkit/core/pytorch/data_util.py
irenaby: There is no reason not to use this for other purposes, as there is nothing hessian-specific here. That's exactly the reason it's here and not under hessian :)

model_compression_toolkit/core/pytorch/data_util.py
irenaby: For fixed dataset shuffling can be defined in dataloader as it's a regular torch Dataset. 
For iterable dataset shuffling is problematic and torch doesn't allow it, so the provided data generator should take care of that. This is exactly the same as today. Iterable dataset follows the data generator precisely, I think its mentioned in the documentation

model_compression_toolkit/gptq/common/gptq_training.py
irenaby: I didn't write this line!

model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation.py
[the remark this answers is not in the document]
ofirgo: Let me just verify that I understand correctly—the "user" (meaning us) can pass a Python generator, and it'll be okay?What do you mean by "if the batch size was appropriate"?
It just seems like complicated requirement for the user, that exposes internal implementation mechanism of the service and requires the user to deal with it.
If you say that the request excepts a "simple" python generator and here it is built as a loader just to define a different batch size then it's ok

model_compression_toolkit/core/common/pruning/importance_metrics/lfh_importance_metric.py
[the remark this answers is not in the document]
ofirgo: Yes you are correct that it should be identical, and also that the small change due to randomness is probably negligible, but I prefer to avoid changing unrelated algorithms to prevent unexpected issues, this is already a big enough refactor that affects many modules

model_compression_toolkit/gptq/common/gptq_training.py
ofirgo: I know, but you also changed many other such exceptions in places you modified other code (like in the fw implementation) so I just wondered why here you decided to let it go 😄

tests/keras_tests/function_tests/test_hmse_error_method.py
```diff
@@ -110,52 +109,58 @@ def _run_node_verification(node_type):
...
+                                                            data_loader=None,
+                                                            n_samples=1,
                                                             target_nodes=[node])
-
-            self.assertTrue(self.his.count_saved_scores_of_request(expected_hessian_request)[node] > 0,
-                            f"No Hessian-based scores were computed for node {node}, "
-                            "but expected parameters selection to run with HMSE.")
+            # check hessians have been precomputed (dataloader=None fetches from cache)
+            hess = self.his.fetch_hessian(expected_hessian_request)
+            self.assertTrue(hess[node.name].shape[0] == 1)
```
ofirgo, line 119: add a message to this assertion, so it would be simpler to understand what was the expected behavior (None dataloader should fetch from cache) in case it fails

tests/pytorch_tests/function_tests/test_hessian_info_calculator.py
```diff
@@ -35,7 +36,7 @@
 class basic_model(torch.nn.Module):
     def __init__(self):
         super(basic_model, self).__init__()
-        self.conv1 = Conv2d(3, 3, kernel_size=1, stride=1)
+        self.conv1 = Conv2d(3, 3, kernel_size=3, stride=1, padding=1)
```
ofirgo, line 39: why is this change necessary?

ofirgo (changes requested):

tests/pytorch_tests/function_tests/test_hessian_info_calculator.py
irenaby: Small kernel with small input randomly zero the gradients (I assume relu randomly gets all-negative inputs). This was the simplest "solution" to significantly decrease the chance of that happening.

model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation.py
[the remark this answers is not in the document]
irenaby: Let's talk

model_compression_toolkit/gptq/common/gptq_training.py
irenaby: I didn't decide, I just didn't notice it. Probably losing sight due to old age, that's the only possible explanation

ofirgo: that's what I feared of 😆

model_compression_toolkit/gptq/pytorch/gptq_training.py
irenaby: Not sure if there was an issue here or I just didn't want to waste time on something that currently works as I was instructed :)  Most probably it's the last one

model_compression_toolkit/core/common/hessian/hessian_info_service.py
```diff
@@ -12,19 +12,91 @@
...
+        Returns:
+            Minimal samples count after update (among updated layers).
+
+        """
+        assert set(layers_hessians.keys()) == set(n.name for n in request.target_nodes)
+        n_nodes_samples = []   # samples count per node after update
+        for node_name, hess in layers_hessians.items():
+            query = Query(request.mode, request.granularity, node_name)
+            saved_hess = self._data.get(query)
+            new_hess = hess if saved_hess is None else np.unique(np.concatenate([saved_hess, hess], axis=0), axis=0)
```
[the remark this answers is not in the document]
irenaby: Yeah, calculators return the results as np. But you are right, I failed to keep it tensor-type agnostic, will change.

model_compression_toolkit/core/common/pruning/importance_metrics/lfh_importance_metric.py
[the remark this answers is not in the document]
irenaby: restored

model_compression_toolkit/core/common/hessian/hessian_info_service.py
```diff
@@ -12,21 +12,96 @@
...
+    granularity: HessianScoresGranularity
+    node: LayerName
+
+
+class HessianCache:
+    """ Hessian cache """
+    def __init__(self):
+        self._data: Dict[Query, Tensor] = {}
+
+    def update(self, layers_hessians, request) -> int:
```
ofirgo, line 46: add typehints

ofirgo (approved):

PR #1244, opened 2024-10-14
Gradual activation quantization in Keras

reuvenperetz:
## Pull Request Description:
Add support for gradual activation quantization in Keras.
This mainly converts torch implementation of gradual activation quantization to a common implementation.
In addition, pytests tests were added to keras (similar to current pytorch tests).


model_compression_toolkit/gptq/common/regularization_factory.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 Sony Semiconductor Israel, Inc. All rights reserved.
+# Copyright 2024 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo, line 1: why? this is not a new file

model_compression_toolkit/trainable_infrastructure/common/annealing_schedulers.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        if not (0 <= t_start < t_end):
+            raise ValueError(f'Expected 0 <= t_start < t_end, actual {t_end=} {t_start=}')
+
+        self.t_start = t_start
+        self.t_end = t_end
+        self.initial_val = initial_val
+        self.target_val = target_val
+
+    def _compute_factor(self, t: int) -> float:
```
ofirgo, line 37: is it possible to use `@abstractmethod` here? (not sure because it does not inherit from ABC)

model_compression_toolkit/trainable_infrastructure/common/annealing_schedulers.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        Calculates the annealed value based on the current time step `t`.
+
+        Args:
+            t: Current time step.
+
+        Returns:
+            float: Annealed value between initial_val and target_val.
+        """
+        factor = self._compute_factor(t)
```
ofirgo, line 62: nice! I wish we had more of these in our code

model_compression_toolkit/gptq/keras/quantizer/quantization_builder.py
```diff
@@ -70,12 +73,13 @@ def quantization_builder(n: common.BaseNode,
...
+                                                        quantizer_id=TrainingMethod.STE,
                                                         quant_method=quant_method,
-                                                        quantizer_base_class=BaseKerasInferableQuantizer)
+                                                        quantizer_base_class=BaseKerasActivationTrainableQuantizer)
+        cfg = get_trainable_quantizer_activation_config(n, None)
 
-        kwargs = get_inferable_quantizer_kwargs(n.final_activation_quantization_cfg, QuantizationTarget.Activation)
-
-        activation_quantizers.append(quantizer_class(**kwargs))
+        # freeze_quant_params is True since in GPTQ the quantization parameters should not be trained (previously, we used inferable quantizers).
```
ofirgo, line 82: "the **activation** quantization parameters should not..."
and you can remove the "previously.." part IMO

model_compression_toolkit/gptq/keras/quantization_facade.py
```diff
@@ -94,19 +96,37 @@ def get_keras_gptq_config(n_epochs: int,
...
             >>> gptq_conf = mct.gptq.get_keras_gptq_config(n_epochs=5)
 
             Other Tensorflow optimizers can be passed:
 
             >>> gptq_conf = mct.gptq.get_keras_gptq_config(n_epochs=3, optimizer=tf.keras.optimizers.Nadam())
 
-            The configuration can be passed to :func:`~model_compression_toolkit.keras_post_training_quantization` in order to quantize a keras model using gptq.
+            To enable Gradual Activation Quantization with non-default settings build GradualActivationQuantizationConfig:
+
+            >>> gradual_act_conf = mct.gptq.GradualActivationQuantizationConfig(mct.gptq.QFractionLinearAnnealingConfig(initial_q_fraction=0.2))
```
ofirgo, line 109: this looks very scary, do we really need to specify `initial_q_fraction`?
 even providing the `GradualActivationQuantizationConfig` and not use a default one seems like to much understanding of the algorithm to ask for from the user

model_compression_toolkit/gptq/keras/quantization_facade.py
```diff
@@ -94,19 +96,37 @@ def get_keras_gptq_config(n_epochs: int,
...
+            >>> gptq_conf = mct.gptq.get_keras_gptq_config(n_epochs=3, gradual_activation_quantization=gradual_act_conf)
+
+            The configuration can be passed to :func:`~model_compression_toolkit.keras_gradient_post_training_quantization` in order to quantize a keras model using gptq.
+
 
         """
+        optimizer = optimizer or tf.keras.optimizers.Adam(learning_rate=LR_DEFAULT)
```
ofirgo, line 116: If we are touching this API, lets make sure that this is aligned with the arguments and values of the parallel pytorch API (it might be fine but just lets be sure)

model_compression_toolkit/qat/keras/quantizer/lsq/symmetric_lsq.py
```diff
@@ -29,6 +29,7 @@
 from model_compression_toolkit import constants as C
 
 from model_compression_toolkit.qat.keras.quantizer.base_keras_qat_weight_quantizer import BaseKerasQATWeightTrainableQuantizer
+
```
ofirgo, line 32: unnecessary

model_compression_toolkit/gptq/keras/gptq_training.py
```diff
@@ -77,6 +81,15 @@ def __init__(self,
             hessian_info_service: HessianScoresService for fetching and computing Hessian's approximation scores.
 
         """
+
+        def _get_total_grad_steps():
+            return get_total_grad_steps(representative_data_gen) * gptq_config.n_epochs
+
+        # must be set prior to model building in the base class constructor. TODO: why?
```
ofirgo, line 88: why really?

irenaby: it is used in activation holder construction so it must be set before the model is built, which happens in the base constructor

tests_pytest/keras/gptq/test_gradual_act_quantization.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class TestGradualActivationQuantization:
+
+    def test_gradual_act_quant_wrapper(self, x):
+        quantizer = Quantizer()
+        qw = GradualActivationQuantizerWrapper(quantizer, q_fraction_scheduler=lambda t: t / (t + 1))
+
+        y0, y1, y2 = [qw(x, training=True) for _ in range(3)]
+        np.testing.assert_array_almost_equal(y0.numpy(), x.numpy())  # t=0
```
ofirgo, line 45: Why np.testing? is there any advantages of using this tool?
unless there is a clear advantage I think it is preferred that we'll be aligned with the assertion methodology that Irena used for the Pytorch Pytests (which is, to use Python's built in assertion).

ofirgo (changes requested):

model_compression_toolkit/gptq/keras/quantization_facade.py
reuvenperetz: It's possible to remove this, but we need to demonstrate to the user how to use and configure GradualActivationQuantizationConfig somehow.

Do you think this should be in the GPTQ tutorial instead? Even though I think it should be here since this is what will be generated in the docs pages.

ofirgo: I  do think it should be in a tutorial, and here we should only have a "plug-and-play" example. I think that just giving a line of code that configures something without getting into the meaning of every argument and module that is used can be confusing

model_compression_toolkit/gptq/keras/quantization_facade.py
ofirgo: I have no idea how the values were chosen and why they differ between the frameworks. Let's align both defaults to the default value in the accuracy runner (I think its 1e-2 for both lr and reg factor but double check that)

tests_pytest/keras/__init__.py
```diff
@@ -1,14 +1,14 @@
-# Copyright 2024 Sony Semiconductor Israel, Inc. All rights reserved.
+#  Copyright 2024 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo, line 1: remove spaces

ofirgo (approved):

PR #1253, opened 2024-10-27
fix hmse overridden by mse after resource utilization computation

irenaby:
## Pull Request Description:
When GPTQ with HMSE follows resource utilization computation, it runs with MSE instead of HMSE. 
Resource utilization computation runs graph_preparation_runner with default gptq flag which is false. In set_quantization_configuration_to_graph when weights_error_method is HMSE and gptq is false, weights_error_method is set to MSE in the original core_config. So when gptq is run afterwards, its config contains MSE instead of HMSE.

This fix runs resource utilization on a copy of core config, so that the original config is not changed. 

### DIFF
--- a/model_compression_toolkit/core/common/mixed_precision/resource_utilization_tools/resource_utilization_data.py
+++ b/model_compression_toolkit/core/common/mixed_precision/resource_utilization_tools/resource_utilization_data.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import copy
+
 import numpy as np
 from typing import Callable, Any, Dict, Tuple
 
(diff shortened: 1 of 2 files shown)


irenaby:
@ofirgo I'm not sure this is the best fix, opening PR so I don't forget, let's talk on Tue

model_compression_toolkit/core/common/mixed_precision/resource_utilization_tools/resource_utilization_data.py
```diff
@@ -247,3 +251,20 @@ def requires_mixed_precision(in_model: Any,
...
+    Create a core config to use for resource utilization computation.
+
+    Args:
+        core_config: input core config
+
+    Returns:
+        Core config for resource utilization.
+    """
+    core_config = copy.deepcopy(core_config)
+    # HMSE is not supported without GPTQ, and resource utilization doesn't care about it, so we replace it with MSE.
```
lior-dikstein, line 267: For resource utilization, we use GPTQ disabled, and since HMSE is not supported without GPTQ we replace it with MSE.

lior-dikstein (approved):

PR #1257, opened 2024-11-04
fix package discovery for wheel

irenaby:
## Pull Request Description:
tests_pytest is included in the wheel as it was not explicitly excluded from find_packages.
Use include instead of exclude in find_packages to automatically exclude any new non-product code.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,10 +42,7 @@ setup(name='model_compression_toolkit',
       long_description=get_log_description(),
       long_description_content_type="text/markdown",
       description='A Model Compression Toolkit for neural networks',
-      packages=find_packages(
-          exclude=["tests", "tests.*",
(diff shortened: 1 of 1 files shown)


ofirgo (commented):
What about the docs? I don't know if this should be included as part of the package, but it wasn't excluded before.
I prefer to leave it with the exclude and just add the new package that we want to exclude

ofirgo:
> @ofirgo docs are not included in the package. find_packages only looks for python packages, anything else need to be included explicitly with other arguments (i also don't see any reason why docs would be included in the wheel). Including only the product src code is much more robust, it was pure coincidence that I notice this at all.

Ok, sounds reasonable

ofirgo (approved):

PR #1276, opened 2024-11-24
Refactor Target Platform Capabilities Design

lior-dikstein:
- Create a new `schema` package to house all target platform modeling classes
- Introduce a new versioning system with minor and patch versions

Additional Changes:
- Update existing target platform models to adhere to the new versioning convention
- Add necessary metadata
- Correct all import statements
- Update and enhance tests to reflect the design changes


model_compression_toolkit/metadata.py
```diff
@@ -43,13 +43,16 @@ def create_model_metadata(tpc: TargetPlatformCapabilities,
...
 
-    Returns: A dictionary with TPC and MCT versions.
+    Returns: A dictionary with TPC, MCT and TPC-Schema versions.
 
     """
     # imported inside to avoid circular import error
     from model_compression_toolkit import __version__ as mct_version
-    tpc_version = f'{tpc.name}.{tpc.version}'
-    return {MCT_VERSION: mct_version, TPC_VERSION: tpc_version}
+    tpc_version = f'{tpc.tp_model.tpc_minor_version}.{tpc.tp_model.tpc_patch_version}'
```
ofirgo, line 51: I know we said to concat the minor and patch, but on second thought, it would be better to just save to separate fields

model_compression_toolkit/target_platform_capabilities/schema/v1.py
```diff
@@ -0,0 +1,727 @@
...
+            tpc_patch_version (Optional[int]): The patch version of the target platform capabilities.
+            add_metadata (bool): Whether to add metadata to the model or not.
+            name (str): Name of the model.
+
+         Raises:
+            AssertionError: If the provided `default_qco` does not contain exactly one quantization configuration.
+        """
+
+        super().__init__()
+        self.tpc_minor_version = tpc_minor_version
```
ofirgo, line 555: need to add another foeld to the versioning scheme - "platform_type" (e.g., imx500, aisoc...).
make it a string and add it to the metadata as well

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -22,8 +22,9 @@
     ACTIVATION_N_BITS_ATTRIBUTE, FP32_BYTES_PER_PARAMETER
 from model_compression_toolkit.core.common.quantization.node_quantization_config import WeightsAttrQuantizationConfig
 from model_compression_toolkit.logger import Logger
-from model_compression_toolkit.target_platform_capabilities.target_platform import QuantizationConfigOptions, \
-    TargetPlatformCapabilities, LayerFilterParams, OpQuantizationConfig
+from model_compression_toolkit.target_platform_capabilities.target_platform import TargetPlatformCapabilities, LayerFilterParams
+from model_compression_toolkit.target_platform_capabilities.schema.v1 import OpQuantizationConfig, \
```
ofirgo, line 26: Given that the schema is going to be updated (maybe frequently), and MCT core would change the schema that it is using accordingly, we need to figure out how to modify these imports throughout the code without accessing ".v1" directly at each import.

Is there a way to "export" (as TPC package "API") a default schema that references the currently used version, such that all imports will point to it and when we want to change the used schema by MCT we'll only have to change in 1 place?

Maybe @irenaby would have an idea how it can be done?

Let's discuss this offline if needed.

ofirgo: BTW, this is not mandatory for this PR, but solving this here would be better, because it will save us editing all these files again in a separate PR

model_compression_toolkit/target_platform_capabilities/target_platform/fusing.py
```diff
@@ -14,72 +14,3 @@
 # ==============================================================================
 
 
-from typing import Any, List, Union
```
ofirgo, line 17: empy file, you can remove it

model_compression_toolkit/target_platform_capabilities/target_platform/op_quantization_config.py
```diff
@@ -13,366 +13,4 @@
 # limitations under the License.
 # ==============================================================================
 
-import copy
```
ofirgo, line 16: empty file

model_compression_toolkit/target_platform_capabilities/target_platform/operators.py
```diff
@@ -12,76 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-from typing import Dict, Any
```
ofirgo, line 15: empty file

model_compression_toolkit/target_platform_capabilities/target_platform/target_platform_model.py
```diff
@@ -13,19 +13,8 @@
 # limitations under the License.
 # ==============================================================================
 
-import pprint
```
ofirgo, line 16: empty file

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/target_platform_capabilities.py
```diff
@@ -24,40 +24,30 @@
...
-from model_compression_toolkit.constants import MCT_VERSION, TPC_VERSION
 
 
 class TargetPlatformCapabilities(ImmutableClass):
     """
     Attach framework information to a modeled hardware.
     """
-    def __init__(self,
-                 tp_model: TargetPlatformModel,
-                 name: str = "base",
```
ofirgo, line 41: I don't think that we need to remove the name here (just the version)

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v1/tp_model.py
```diff
@@ -102,7 +102,7 @@ def get_op_quantization_configs() -> Tuple[OpQuantizationConfig, List[OpQuantiza
         signedness=Signedness.AUTO)
 
     # We define an 8-bit config for linear operations quantization, that include a kernel and bias attributes.
-    linear_eight_bits = tp.OpQuantizationConfig(
+    linear_eight_bits = model_compression_toolkit.target_platform_capabilities.schema.v1.OpQuantizationConfig(
```
ofirgo, line 105: maybe import the path as "schema_v1" (or create an alias at the beginning of the file) instead of repeating it everywhere?
or do we want it to be explicit in the TPC model description? @haihabi 
even if we want it to, eventually this file will be a JSON and not written as code using imports, so these fields in the JSON won't include the entire schema path anyway...

tests/common_tests/test_tp_model.py
```diff
@@ -25,7 +26,7 @@
 tp = mct.target_platform
 
 TEST_QC = generate_test_op_qc(**generate_test_attr_configs())
-TEST_QCO = tp.QuantizationConfigOptions([TEST_QC])
+TEST_QCO = model_compression_toolkit.target_platform_capabilities.schema.v1.QuantizationConfigOptions([TEST_QC])
```
ofirgo, line 29: I think that at least in the tests you should definitely use an alias for this path.
Also, a reference to a "current schema" pointer is very important for the tests as well (we want the tests to run with the active schema used by MCT without needing to change all this imports each time we update a schema)

model_compression_toolkit/core/common/graph/base_node.py
irenaby: The only part of MCT that should be aware of the schema is the parser once we have it. It should parse the schema into whatever representation the rest of mct works with. If we want to reuse the same classes for now, we can add a proxy module that will only import the classes from the schema, and the rest of mct imports from that proxy module.

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v1/tp_model.py
haihabi: But the JSON file should have the schema version from the constant in the schema main class.

model_compression_toolkit/core/common/graph/base_node.py
haihabi: I agree with @irenaby , we should have some proxy model at this stage...

model_compression_toolkit/metadata.py
```diff
@@ -43,13 +45,20 @@ def create_model_metadata(tpc: TargetPlatformCapabilities,
...
     """
     # imported inside to avoid circular import error
     from model_compression_toolkit import __version__ as mct_version
-    tpc_version = f'{tpc.name}.{tpc.version}'
-    return {MCT_VERSION: mct_version, TPC_VERSION: tpc_version}
+    tpc_minor_version = f'{tpc.tp_model.tpc_minor_version}'
+    tpc_patch_version = f'{tpc.tp_model.tpc_patch_version}'
+    tpc_platform_type = f'{tpc.tp_model.tpc_platform_type}'
+    tpc_schema = f'{tpc.tp_model.SCHEMA_VERSION}'
+    return {MCT_VERSION: mct_version,
```
irenaby, line 57: If you need to access the fields elsewhere, why not define class or named tuple? If this goes directly into the model, no need to define global consts.

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/v1/tp_model.py
ofirgo: > But the JSON file should have the schema version from the constant in the schema main class.

So you agree? the path here can be shortened with an alias to improve file readability?

haihabi: Yes

model_compression_toolkit/qat/keras/quantizer/README.md
```diff
@@ -5,7 +5,7 @@
 Several training methods may be applied by the user to train the QAT ready model
 created by `keras_quantization_aware_training_init` method in [`keras/quantization_facade`](../quantization_facade.py).
 Each `TrainingMethod` (an enum defined in the [`qat_config`](../../common/qat_config.py)) 
-and [`QuantizationMethod`](../../../target_platform_capabilities/target_platform/op_quantization_config.py)
+and `QuantizationMethod`
```
ofirgo, line 8: fix the path according to the new file location

ofirgo (approved):

PR #1277, opened 2024-11-26
Add SLA in Keras

reuvenperetz:
## Pull Request Description:
Add support for SLA in Keras.


model_compression_toolkit/core/keras/data_util.py
```diff
@@ -29,39 +50,151 @@ def flat_gen_fn(data_gen_fn: Callable[[], Generator]):
...
+        self.samples = samples
+
+    def __len__(self):
+        return len(self.samples)
+
+    def __getitem__(self, index):
+        return self.samples[index]
+
+
+class FixedSampleInfoDataset:
```
irenaby, line 127: if it's kept, lets align with torch api

model_compression_toolkit/core/keras/data_util.py
```diff
@@ -29,39 +50,151 @@ def flat_gen_fn(data_gen_fn: Callable[[], Generator]):
...
             for sample in zip(*inputs_batch):
-                yield to_tf_tensor(sample)
-    return gen
+                yield tuple([tf.convert_to_tensor(s) for s in sample])
 
+    return gen
 
-# TODO in tf dataset and dataloader are combined within tf.data.Dataset. For advanced use cases such as gptq sla we
-#  need to separate dataset from dataloader similarly to torch data_util.
 class TFDatasetFromGenerator:
```
irenaby, line 61: I think trying to build exactly the same objects as in torch is more confusing than helpful. In torch Dataset is indended to be extended, and that's what every custom dataset class does. Tf isn't built the same, and here some "dataset" classes contain tf dataset, and some not, and then for dataloader you convert the dataset back into generator only to then convert back into dataset that serves as dataloader. I suggest we don't try to imitate torch

reuvenperetz: I agree this seems a bit weird, but as you wrote above, "Let's try to be consistent in the logic between frameworks, one way or another." The only reason this is built this way is to make sure the frameworks are aligned, and I do not see a different implementation that can be similar to torch, yet very different from the current implementation (except points you mentioned like FixedSampleInfoDataset and extracting common logic to base class).

model_compression_toolkit/core/keras/hessian/activation_hessian_scores_calculator_keras.py
```diff
@@ -60,96 +60,115 @@ def compute(self) -> List[np.ndarray]:
...
                             new_mean_res = tf.reduce_mean(tf.stack(ipts_hessian_approximations), axis=1)
                             relative_delta_per_node = (tf.abs(new_mean_res - prev_mean_results) /
                                                        (tf.abs(new_mean_res) + 1e-6))
                             max_delta = tf.reduce_max(relative_delta_per_node)
-                            if max_delta < HESSIAN_COMP_TOLERANCE:
-                                break
+
+                        elif self.hessian_request.granularity == HessianScoresGranularity.PER_ELEMENT:
+                            # Reshape to maintain spatial dimensions and average only across batch
+                            new_mean_res = [tf.reduce_mean(h, axis=0) for h in ipts_hessian_approximations]
```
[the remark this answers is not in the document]
reuvenperetz: it was removed

irenaby (approved):

PR #1297, opened 2024-12-18
Refactor Target Platform Capabilities - Phase 3

lior-dikstein:
Remove context manager functionality from Target Platform Model. Fix all tests and tp models.
Make all schema classes fully immutable (no more lists).


model_compression_toolkit/target_platform_capabilities/schema/schema_functions.py
```diff
@@ -82,7 +82,10 @@ def is_opset_in_model(tp_model: TargetPlatformModel, opset_name: str) -> bool:
         bool: True if an OperatorsSet with the given name exists in the target platform model,
               otherwise False.
     """
-    return opset_name in [x.name for x in tp_model.operator_set]
```
ofirgo, line 85: `return tp_model.operator_set is not None and opset_name in [x.name for x in tp_model.operator_set]`

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/target_platform_capabilities.py
```diff
@@ -52,7 +52,7 @@ def __init__(self,
         self.op_sets_to_layers = OperationsToLayers() # Init an empty OperationsToLayers
         self.layer2qco, self.filterlayer2qco = {}, {} # Init empty mappings from layers/LayerFilterParams to QC options
         # Track the unused opsets for warning purposes.
-        self.__tp_model_opsets_not_used = [s.name for s in tp_model.operator_set]
+        self.__tp_model_opsets_not_used = [s.name for s in tp_model.operator_set] if tp_model.operator_set else []
```
ofirgo, line 55: Can tp_model.operator_set be None? isn't it something that we want to raise exception on or is it a common case that we can handle?
anyway, I don;t like the empty list initialization here, if a None tp_model.operator_set might raise issues going forward than we should address it and not hide it with an empty list

ofirgo (changes requested):

ofirgo (approved):

PR #1308, opened 2025-01-01
Use Attach TPC to Framework as part of MCT main track

ofirgo:
## Pull Request Description:

This PR changes the TPC mechanism in MCT APIs.
All the main API functions now get a framework-agnostic target platform description and use the Attach2Fw module to generate a framework-dependent quantization capabilities description.

The main changes include:
- Alignment of the existing TP models with the new schema (v1).
- Remove TP models v2-4 for IMX500 (moved to tests for simplicity of testing).
- Remove existing fw TPCs (no need for them anymore).
- Addition of attach2fw usage in all API functions.
  - To prevent breaking changes for users using the default TPC, the `get_target_platform_capabilities` API function is kept with a reduced functionality that allows to only obtain the default v1 TPC.
- Minor changes to the built-in OperatorSetsNames enum to support more required operators.
- Massive tests modification and alignment to comply with the new TPC mechanism.


ofirgo:
@lior-dikstein @irenaby 
Please start going over the main code and API changes.
Note that the work on the tests is still not done, so don't go over the tests yet.
Thanks

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/operations_to_layers.py
```diff
@@ -138,9 +138,7 @@ def validate_op_sets(self):
                               OperationsSetToLayers), f'Operators set should be of type OperationsSetToLayers but it ' \
                                                       f'is of type {type(ops2layers)}'
 
-            # Assert that opset in the current TargetPlatformCapabilities and has a unique name.
```
ofirgo, line 141: Note that this assertion is moved to the attach to fw module

model_compression_toolkit/target_platform_capabilities/tpc_models/imx500_tpc/latest/__init__.py
```diff
@@ -16,9 +16,8 @@
 from model_compression_toolkit.target_platform_capabilities.tpc_models.imx500_tpc.v1.tp_model import get_tp_model, generate_tp_model, \
     get_op_quantization_configs
 if FOUND_TF:
-    from model_compression_toolkit.target_platform_capabilities.tpc_models.imx500_tpc.v1.tpc_keras import get_keras_tpc as get_keras_tpc_latest
-    from model_compression_toolkit.target_platform_capabilities.tpc_models.imx500_tpc.v1.tpc_keras import generate_keras_tpc
+    from model_compression_toolkit.target_platform_capabilities.tpc_models.imx500_tpc.v1.tp_model import get_tp_model as get_keras_tpc_latest
```
ofirgo, line 19: I'm aware that this is not the best way to expose this APIs, but note that this change is done in all /latest/ tpc init files, to support the current default API and testing.

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2keras.py
```diff
@@ -76,27 +77,47 @@ def __init__(self):
...
+            OperatorSetNames.OPSET_SOFTMAX.value: [tf.nn.softmax, Softmax,
+                                                   LayerFilterParams(Activation, activation="softmax")],
+            OperatorSetNames.OPSET_LOG_SOFTMAX.value: [tf.nn.log_softmax],
+            OperatorSetNames.OPSET_ADD_BIAS.value: [tf.nn.bias_add],
+            OperatorSetNames.OPSET_L2NORM.value: [tf.math.l2_normalize],
         }
 
         if FOUND_SONY_CUSTOM_LAYERS:
-            self._opset2layer[OperatorSetNames.OPSET_POST_PROCESS] = [SSDPostProcess]
+            self._opset2layer[OperatorSetNames.OPSET_SSD_POST_PROCESS.value] = [SSDPostProcess]
```
irenaby, line 108: Why are there two opsets for the same layer?

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2pytorch.py
```diff
@@ -18,28 +18,28 @@
...
 
 
-class AttachTpModelToPytorch(AttachTpModelToFw):
+class AttachTpcToPytorch(AttachTpcToFramework):
     def __init__(self):
         super().__init__()
 
         self._opset2layer = {
             OperatorSetNames.OPSET_CONV.value: [Conv2d],
+            OperatorSetNames.OPSET_DEPTHWISE_CONV.value: [],  # no specific operator for depthwise conv in pytorch
```
irenaby, line 42: Why do we need a separate opset for depthwise at all? Regular conv can be a depthwise, they must have the same quantization configuration, don't they?

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/operations_to_layers.py
```diff
@@ -138,9 +138,7 @@ def validate_op_sets(self):
...
-            opset_in_model = is_opset_in_model(_current_tpc.get().tp_model, ops2layers.name)
-            assert opset_in_model, f'{ops2layers.name} is not defined in the target platform model that is associated with the target platform capabilities.'
+            # Assert that opset has a unique name.
             assert not (ops2layers.name in existing_opset_names), f'OperationsSetToLayers names should be unique, but {ops2layers.name} appears to violate it.'
```
irenaby, line 142: Its more idiomatic to use ops2layers.name not in

model_compression_toolkit/core/common/quantization/quantization_config.py
```diff
@@ -86,6 +88,7 @@ class QuantizationConfig:
     concat_threshold_update: bool = False
     activation_bias_correction: bool = False
     activation_bias_correction_threshold: float = 0.0
+    custom_tpc_opset_to_layer: Optional[Dict[str, Tuple[List[Any], Optional[Dict[str, DefaultDict]]]]] = None
```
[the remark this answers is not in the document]
irenaby: More importantly, what is it for? Why QuantizationConfig should be aware of anything custom? By the time tpc model is consumed, I would expect no difference between built in and custom operators.

model_compression_toolkit/target_platform_capabilities/tpc_models/get_target_platform_capabilities.py
```diff
@@ -12,45 +12,63 @@
...
+    This is a degenerated function that only returns the MCT default TargetPlatformModel object, to comply with the
+    existing TPC API.
 
     Args:
-        fw_name: Framework name of the TargetPlatformCapabilities.
-        target_platform_name: Target platform model name the model will use for inference.
-        target_platform_version: Target platform capabilities version.
+        fw_name: Framework name of the TargetPlatformCapabilities (not used in this function).
+        target_platform_name: Target platform model name the model will use for inference (not used in this function).
+        target_platform_version: Target platform capabilities version (not used in this function).
```
lior-dikstein, line 37: why do you write "not used in this function"?

model_compression_toolkit/xquant/keras/keras_report_utils.py
```diff
@@ -40,10 +43,15 @@ def __init__(self, report_dir: str):
         fw_info = DEFAULT_KERAS_INFO
         fw_impl = KerasImplementation()
 
+        # Set the default Target Platform Capabilities (TPC) for PyTorch.
```
lior-dikstein, line 46: replace "Pytorch" with "Keras"

tests/common_tests/helpers/tpcs_for_tests/v1/tp_model.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024 Sony Semiconductor Israel, Inc. All rights reserved.
+# Copyright 2023 Sony Semiconductor Israel, Inc. All rights reserved.
```
lior-dikstein, line 1: ?

tests/common_tests/helpers/tpcs_for_tests/v1/tp_model.py
```diff
@@ -45,8 +43,7 @@ def get_tp_model() -> TargetPlatformModel:
                              name='imx500_tp_model')
 
 
-def get_op_quantization_configs() -> \
-        Tuple[OpQuantizationConfig, List[OpQuantizationConfig], OpQuantizationConfig]:
+def get_op_quantization_configs() -> Tuple[OpQuantizationConfig, List[OpQuantizationConfig], OpQuantizationConfig]:
```
lior-dikstein, line 46: generate_default_op_quantization_config

tests/keras_tests/feature_networks_tests/feature_networks/mixed_precision_bops_test.py
```diff
@@ -38,6 +39,10 @@ def __init__(self, unit_test, mixed_precision_candidates_list):
 
         self.mixed_precision_candidates_list = mixed_precision_candidates_list
 
+    def get_core_config(self):
+        return CoreConfig(quantization_config=QuantizationConfig(
```
lior-dikstein, line 43: I don't know if you want to change this, but "get" should be "generate" because you generate a core  config, you don't fetch it from a list or dict

tests/keras_tests/function_tests/test_hmse_error_method.py
```diff
@@ -69,14 +71,21 @@ def get_tpc(quant_method, per_channel):
     tp = generate_test_tp_model(edit_params_dict={
         'weights_quantization_method': quant_method,
         'weights_per_channel_threshold': per_channel})
-    tpc = generate_keras_tpc(name="hmse_params_selection_test", tp_model=tp)
 
-    return tpc
+    return tp
```
lior-dikstein, line 75: tp_model?

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2fw.py
```diff
@@ -35,22 +37,30 @@ def attach(self, tpc_model: TargetPlatformModel,
         """
 
         tpc = TargetPlatformCapabilities(tpc_model)
+        custom_opset2layer = custom_opset2layer if custom_opset2layer is not None else {}
 
         with tpc:
```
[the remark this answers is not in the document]
ofirgo: We got rid of the context manager for the TargetPlatformModel class, because it was affecting the schema definition.
We currently kept it for the TargetPlatformCapabilities class to prevent additional changes that might lead to an even larger refactor, which is unnecessary at this point.
We might remove it from this part in the future.

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2keras.py
```diff
@@ -76,27 +77,47 @@ def __init__(self):
...
+            # If Custom layers is not installed then we don't want the user to fail, but just ignore custom layers
+            # in the initialized framework TPC
+            self._opset2layer[OperatorSetNames.OPSET_SSD_POST_PROCESS.value] = []
 
-        self._opset2attr_mapping = {OperatorSetNames.OPSET_CONV.value: {
-            KERNEL_ATTR: DefaultDict(default_value=KERAS_KERNEL),
-            BIAS_ATTR: DefaultDict(default_value=BIAS)},
+        self._opset2attr_mapping = {
+            OperatorSetNames.OPSET_CONV.value: {
+                KERNEL_ATTR: DefaultDict(default_value=KERAS_KERNEL),
```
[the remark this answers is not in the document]
ofirgo: are you asking why defaultdict and not simple dict or what is the purpose of the dict when setting an opset to layer mapping?
if the former - to eliminate the need for defining an entry for every layer type in the opset (usually they all have the same attribute name)
if the latter - to support attributes quantization beside the kernel. when converting to a few dependent TPC we need to get the fw name of the attributes.

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2keras.py
```diff
@@ -53,10 +53,10 @@ def __init__(self):
             OperatorSetNames.OPSET_UNSTACK.value: [tf.unstack],
```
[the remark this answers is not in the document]
ofirgo: I agree and open for suggestions

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2keras.py
ofirgo: why two? I see only OPSET_SSD_POST_PROCESS

model_compression_toolkit/target_platform_capabilities/schema/v1.py
```diff
@@ -33,9 +34,11 @@ class OperatorSetNames(Enum):
     OPSET_GATHER = "Gather"
```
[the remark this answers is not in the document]
ofirgo: I agree, let's take care of this as part of the naming refactor (after this PR).

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2pytorch.py
ofirgo: in keras there is a separate layer type specifically for depthwise.
and not necessarily, it is possible that you would want to quantize a depthwise conv in a certain way and a general conv in a different way (for instance if you know that the depthwise are a bottleneck in your model and you want to keep them in higher precision)

model_compression_toolkit/core/keras/resource_utilization_data_facade.py
```diff
@@ -78,6 +81,12 @@ def keras_resource_utilization_data(in_model: Model,
 
         fw_impl = KerasImplementation()
 
+        # Attach tpc model to framework
+        attach2keras = AttachTpcToKeras()
```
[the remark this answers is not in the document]
ofirgo: I don't really care, maybe just because the object class name is "AttachTo.." I would prefer to call the variable with a similar name, but I can be persuaded otherwise if you think it is critical

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2fw.py
```diff
@@ -35,22 +37,30 @@ def attach(self, tpc_model: TargetPlatformModel,
...
+            for opset in tpc_model.operator_set:
+                if isinstance(opset, OperatorsSet):  # filter out OperatorsSetConcat
+                    if opset.name in custom_opset2layer:
+                        operators = custom_opset2layer[opset.name]
+                        if len(operators) == 1:
+                            OperationsSetToLayers(opset.name, operators[0])
+                        elif len(operators) == 2:
+                            OperationsSetToLayers(opset.name, operators[0], attr_mapping=operators[1])
+                        else:
+                            raise ValueError(f"Custom operator set to layer mapping should include up to 2 elements - "
```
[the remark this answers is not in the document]
ofirgo: Let's discuss this offline (I don't really have a good answer besides that I want to reduce our usage of the Logger until we take care of the way we handle errors)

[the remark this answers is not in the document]
ofirgo: I wouldn't want to go over the dictionary and verify it until I'm using it.
Moreover, I'll need the if-else statement here anyway to extract the attr_mapping if needed, so why do it twice?

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2pytorch.py
```diff
@@ -57,6 +57,7 @@ def __init__(self):
             OperatorSetNames.OPSET_SUB.value: [operator.sub, sub, subtract],
             OperatorSetNames.OPSET_MUL.value: [operator.mul, mul, multiply],
             OperatorSetNames.OPSET_DIV.value: [operator.truediv, div, divide],
+            OperatorSetNames.OPSET_ADD_BIAS.value: [],  # no specific operator for bias_add in pytorch
```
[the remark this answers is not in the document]
ofirgo: Not sure I understand your suggestion, but what you said is the exact reason for the empty list - we have a TP model that sets an opset (OPSET_ADD_BIAS for instance), but there is no relevant operator in one of the frameworks for this opset. Since I want to enforce that each opset defined in the TP model would have a matching opset mapping in the fw TPC, I have to define something in our hardcoded attach to fw mapping.
If you have a more elegant way of setting this functionality I'll be happy to hear it because I don't like my solution

tests/common_tests/helpers/tpcs_for_tests/v1/tp_model.py
ofirgo: I don't want to change the name of this function this would break to many tests and I don't think that naming is that important here

tests/keras_tests/feature_networks_tests/feature_networks/mixed_precision_bops_test.py
ofirgo: not as part of this PR (hopefully we'll get to a global tests refactor soon)

model_compression_toolkit/core/common/quantization/quantization_config.py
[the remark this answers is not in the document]
ofirgo: > More importantly, what is it for? Why QuantizationConfig should be aware of anything custom? By the time tpc model is consumed, I would expect no difference between built in and custom operators.

We need to let the user define custom operator sets outside our hardcoded set of operators.
The user can basically define any opset name that he wants in his TP model.
So, we need to also provide him with a way to provide us a mapping between the custom opset name to fw operators.
This is a good question about where to put it. For lack of a better idea, I decided to put it in the QuantizationConfig.
Keep in mind that a user that defines a custom op set is an expert user that built his own TPC, probably with our guidance. In such case we'll be able to guide him to fill in this argument in the QuantizationConfig.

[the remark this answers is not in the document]
ofirgo: > You can define types aliases, e.g. SomeMeaningfullNameT = Tuple[.......] custom...: Optional[Dict[str, SomeMeaningfullNameT]] If will be more readable, and pycharm will still show the actual types in its preview. If it repeats in multiple file, we can create a file of such common type definitions, or define them in the relevant place and import from other places.

I agree, I thought about it but wasn't sure what is the best way to do it, and was counting on you to provide me with one (;
Thanks

model_compression_toolkit/gptq/pytorch/quantization_facade.py
```diff
@@ -209,6 +213,11 @@ def pytorch_gradient_post_training_quantization(model: Module,
 
         fw_impl = GPTQPytorchImplemantation()
 
+        # Attach tpc model to framework
+        attach2pytorch = AttachTpcToPytorch()
```
haihabi, line 217: Why not part of quantization core ?

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2keras.py
irenaby: That's because you are young and see well. Apparently I developed double vision

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2pytorch.py
irenaby: You can define any regular conv to be a depthwise conv (groups=channels). If we might want to quantize depthwise and non-depthwise differently, we need to make this distinction for a regular conv as well, otherwise we end up with different quantization for exactly the same op (dw layer and conv layer in dw mode). So dw opset should be mapped to dw layer and conv layer with groups=channels, and conv opset should be mapped to conv layer with groups!=channels. Assuming we don't want to do this at this point, I think it's better to unite the two opsets.

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2keras.py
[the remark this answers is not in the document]
irenaby: Is it supposed to contain a mapping from framework op to non-default attribute? If so it's not clear and quite confusing why defaultdict is used to define KERNEL_ATTR=KERAS_KERNEL. 
Also python already has defaultdict (too bad DefaultDict is used in a gazillion places).

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2pytorch.py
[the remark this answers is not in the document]
irenaby: Does user have to define the mapping for all frameworks? If that's the case, it shouldn't be. User should only have to provide mapping in the framework he works in.

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2keras.py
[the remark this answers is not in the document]
irenaby: Defining the enum as class OpsetNames(str, Enum) should probably suffice. Such enum can be instantiated from string, compared to a string, and json automatically dumps to value. The only thing it may require is adding instantiating from string in the relevant tp class. In any case, the only interfaces that should  do the conversion are immediate writing and reading, anything else should use the enum

model_compression_toolkit/core/common/quantization/quantization_config.py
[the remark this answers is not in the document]
irenaby: Lets talk. 
But also, if a custom op contains weights, how are they mapped to framework?
Also 2: it's not enough to add an interface, MCT should support it too (e.g. looking for attributes based on tp model, I believe that's not what happens today). Was this support added? If not, maybe we shouldn't expose the interface yet.
Also 3: MCT should eventually use the mapping in attach to framework when it looks for specific layers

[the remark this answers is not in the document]
irenaby: Make op list and attrs dict into a songle NamedTuple

model_compression_toolkit/gptq/pytorch/quantization_facade.py
ofirgo: I don't want to expose it to the user as part of the API.
Do you think that we should?

model_compression_toolkit/core/common/quantization/quantization_config.py
```diff
@@ -13,13 +13,29 @@
...
+from typing import Optional, Dict, NamedTuple, List
 
+from model_compression_toolkit import DefaultDict
 from model_compression_toolkit.constants import MIN_THRESHOLD
 
 
+class CustomOpsetLayers(NamedTuple):
+    """
+    This struct is used to define a set of operators to map to a custom operator set defined in the TPC.
+
```
lior-dikstein, line 28: This struct is used to define a mapping between a set of framework operators to a custom opset name.

model_compression_toolkit/target_platform_capabilities/schema/v1.py
```diff
@@ -13,66 +13,74 @@
...
+    CONCATENATE = "Concatenate"
+    STACK = "Stack"
+    UNSTACK = "Unstack"
+    GATHER = "Gather"
+    EXPAND = "Expend"
+    BATCH_NORM = "BatchNorm"
+    L2NORM = "L2Norm"
+    RELU = "ReLU"
+    RELU6 = "ReLU6"
+    LEAKY_RELU = "LEAKYReLU"
```
lior-dikstein, line 40: why not LeakyReLU?

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2fw.py
```diff
@@ -35,22 +38,30 @@ def attach(self, tpc_model: TargetPlatformModel,
         """
 
         tpc = TargetPlatformCapabilities(tpc_model)
+        custom_opset2layer = custom_opset2layer if custom_opset2layer is not None else {}
```
lior-dikstein, line 41: what is the reason not to define it {} in the function input instead of 'None'?

model_compression_toolkit/target_platform_capabilities/schema/v1.py
ofirgo: I don't know you tell me, you gave the values for this enum 😆 
I'll change it

lior-dikstein (approved):

model_compression_toolkit/target_platform_capabilities/target_platform/targetplatform2framework/attach2fw.py
ofirgo: This is bad practice to set mutable values as default arguments

PR #1319, opened 2025-01-08
Add docs build workflow

reuvenperetz:
## Pull Request Description:
Add docs build workflow.


docsrc/source/api/api_docs/modules/qat_config.rst
```diff
@@ -10,10 +10,7 @@ qat_config Module
 
 TrainingMethod
 ================================
-**Select a QAT training method:**
```
ofirgo, line 13: there are many other docsrc modifications that need to be made for the upcoming release, I'm guessing this change is part of this PR because it was required to pass the build?

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@v4
+      - name: Install Python 3
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install sphinx sphinx-autodoc-typehints torch==2.5.* torchvision tensorflow==2.15.*
```
ofirgo, line 27: do we want maybe to force a certain sphinx version? just for stability...

docsrc/source/api/api_docs/modules/qat_config.rst
reuvenperetz: Exactly. This workflow fails if the docs can not be build (like in this case...)

PR #1321, opened 2025-01-12
Renaming TPC 

lior-dikstein:
## Pull Request Description:
Renaming: 
- TargetPlatformModel->TargetPlatformCapabilities
- TargetPlatformCapabilities->FrameworkQuantizationCapabilities
- tp_model->tpc
- tpc->fqc
- target_platform_capabilities->framework_platform_capabilities
- OperatorSetConcat->OperatorSetGroup


model_compression_toolkit/target_platform_capabilities/constants.py
```diff
@@ -21,10 +21,10 @@
 
 
 # Supported TP models names:
-DEFAULT_TP_MODEL = 'default'
```
ofirgo, line 24: I’m debating this change... this is a breaking change for any user that tried to use the generate_tpc API with IMX500_TP_MODEL
Not sure that we want to break every potential user with this, and we can wait with this to the extraction of the IMX500 TPCs to the external repo.

model_compression_toolkit/target_platform_capabilities/target_platform/__init__.py
```diff
@@ -14,9 +14,9 @@
 # ==============================================================================
 
 from model_compression_toolkit.target_platform_capabilities.target_platform.targetplatform2framework.attribute_filter import AttributeFilter
```
ofirgo, line 16: I want to get rid of the model_compression_toolkit/target_platform_capabilities/target_platform level in the hierarchy, there is no need for it anymore since there is only the /targetplatform2framework/ in it.
Is it too complicated for this huge PR or do you think it would be fine to include this here as well?

ofirgo (changes requested):

ofirgo (approved):

PR #1323, opened 2025-01-13
Refactor FixedSampleInfoDataset to Support Multi-Input Models

reuvenperetz:
## Pull Request Description:
The previous implementation of `FixedSampleInfoDataset` caused issues when working with models that expected multiple inputs with different shapes. Specifically, during inference it would raise a ValueError indicating a mismatch between the expected and received number of input tensors.
Instead of directly converting samples to tensors, we now separate the input tensors by their position in the input tuple, create a dictionary mapping each tensor position to its corresponding data, use tf.data.Dataset.from_tensor_slices with the dictionary structure and map the dictionary outputs back to the expected tuple format.
This preserves the tensor grouping and shapes while allowing different shapes for different input layers.

### DIFF
--- a/model_compression_toolkit/core/keras/data_util.py
+++ b/model_compression_toolkit/core/keras/data_util.py
@@ -134,11 +134,30 @@ class FixedSampleInfoDataset:
         self.samples = samples
         self.sample_info = sample_info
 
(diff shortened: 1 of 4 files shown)


ofirgo (approved):
Please wait for this #1321 PR to be merged before merging this one

PR #1334, opened 2025-01-20
Move maxcut warning so it will only appear once.

elad-c:
## Pull Request Description:
Move maxcut warning so it will only appear once.


ofirgo (approved):
We need to also add the condition to required_mixed_precision to always run MP when any other than weights is restricted

PR #1341, opened 2025-01-26
Add tests for ResourceUtilizationCalculator

irenaby:
## Pull Request Description:
Add unit tests for Resource Utilization Calculator.
Minor updates, mostly validations and errors.


elad-c (approved):

irenaby:
@reuvenperetz please look at the github workflows. If we still need to exclude anything from coverage, let's talk.

tests_pytest/common/core/common/mixed_precision/resource_utilization_tools/test_resource_utilization_calculator.py
```diff
@@ -0,0 +1,947 @@
...
+    op_cfg = OpQuantizationConfig(
+        # canonical names (as 'kernel')
+        attr_weights_configs_mapping=attr_weights_configs_mapping,
+        activation_n_bits=a_nbits,
+        enable_activation_quantization=a_enable,
+        default_weight_attr_config=AttributeQuantizationConfig(weights_n_bits=pos_attr[0],
+                                                               enable_weights_quantization=pos_attr[1]),
+        activation_quantization_method=QuantizationMethod.POWER_OF_TWO,
+        quantization_preserving=False,
+        supported_input_activation_n_bits=[2, 4, 8],
```
ofirgo, line 91: are these values relevant for testing or just a placeholder? because 2 and 4 are not relevant values for "real scenarios" so unless they are used to test some specific behavior, maybe change them to 8, 16 which are the relevant options for this field

tests_pytest/common/core/common/mixed_precision/resource_utilization_tools/test_resource_utilization_calculator.py
```diff
@@ -0,0 +1,947 @@
...
+
+    return qc
+
+
+class DummyLayer:
+    """ Only needed for repr(node) to work. """
+    pass
+
+
+def build_node(name='node', canonical_weights: dict=None, qcs=None, input_shape=(4, 5, 6), output_shape=(4, 5, 6),
```
ofirgo, line 128: It seems that the setup for this test (tpc configs, node configs, graph mock, etc.) is a very good setup for many tests that would require such objects.
Maybe we extract it to a 'util' tests section so it would also be available in other tests?
It's probably better to do it in a separate PR, but maybe we can create a kind of "utils API for tests", otherwise I see a case where all this code is duplicated across many tests without changing anything.

tests_pytest/common/core/common/mixed_precision/resource_utilization_tools/test_resource_utilization_calculator.py
irenaby: They are just a placeholder, but anyway the calculator is not and should not be aware of what real scenarios are. If anything I would change it to some random values, but it's not used so it doesn't really matter

tests_pytest/common/core/common/mixed_precision/resource_utilization_tools/test_resource_utilization_calculator.py
irenaby: My only concern is that as a util it will be constantly extended to be more and more configurable, and become a mess. Also someone can decide to change the functionality a bit thinking it doesn't matter, when some previous tests count on that. We should be careful what is really a common utility, and what is the minimal implementation for a specific test.

PR #1347, opened 2025-01-28
Adding integration tests package with ptq e2e test example and converting symmetric selection unit test to pytest

ofirgo:

### DIFF
--- a/tests_pytest/keras_tests/e2e_tests/post_training_quantization_tests/__init__.py
+++ b/tests_pytest/keras_tests/e2e_tests/post_training_quantization_tests/__init__.py
@@ -1,0 +1,14 @@
+# Copyright 2025 Sony Semiconductor Israel, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
--- a/tests_pytest/keras_tests/e2e_tests/post_training_quantization_tests/test_quantization_methods.py
+++ b/tests_pytest/keras_tests/e2e_tests/post_training_quantization_tests/test_quantization_methods.py
@@ -30,22 +30,12 @@ from model_compression_toolkit.target_platform_capabilities.constants import KER
 from tests.common_tests.helpers.tpcs_for_tests.v4.tpc import generate_tpc
 
 INPUT_SHAPE = (224, 224, 3)
-
-
-@pytest.fixture
-def rep_data_gen():
-    np.random.seed(42)
-
-    def representative_dataset():
-        for _ in range(2):
-            yield [np.random.randn(2, *INPUT_SHAPE)]
-
-    return representative_dataset
+NUM_CHANNELS = {layers.Conv2D: 2, layers.Dense: 10}
 
 
 def model_basic():
     inputs = layers.Input(shape=INPUT_SHAPE)
-    x = layers.Conv2D(2, 3, padding='same')(inputs)
+    x = layers.Conv2D(NUM_CHANNELS[layers.Conv2D], 3, padding='same')(inputs)
     x = layers.BatchNormalization()(x)
     x = layers.Activation('relu')(x)
     return tf.keras.models.Model(inputs=inputs, outputs=x)
@@ -53,22 +43,22 @@ def model_basic():
 
 def model_residual():
     inputs = layers.Input(shape=INPUT_SHAPE)
-    x1 = layers.Conv2D(2, 3, padding='same')(inputs)
+    x1 = layers.Conv2D(NUM_CHANNELS[layers.Conv2D], 3, padding='same')(inputs)
     x1 = layers.ReLU()(x1)
 
-    x2 = layers.Conv2D(2, 3, padding='same')(x1)
+    x2 = layers.Conv2D(NUM_CHANNELS[layers.Conv2D], 3, padding='same')(x1)
     x2 = layers.BatchNormalization()(x2)
     x2 = layers.ReLU()(x2)
 
     x = layers.Add()([x1, x2])
 
     x = layers.Flatten()(x)
-    x = layers.Dense(units=10, activation='softmax')(x)
+    x = layers.Dense(units=NUM_CHANNELS[layers.Dense], activation='softmax')(x)
 
     return keras.Model(inputs=inputs, outputs=x)
 
 
-def set_tpc(weights_quantization_method, per_channel):
+def _get_tpc(weights_quantization_method, per_channel):
     # TODO: currently, running E2E test with IMX500 V4 TPC from tests package
     #  we need to select a default TPC for tests, which is the one we want to verify e2e for.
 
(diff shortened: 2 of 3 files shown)


tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+@pytest.fixture
+def tpc_factory():
+    def _tpc_factory(quant_method, per_channel):
+        return set_tpc(quant_method, per_channel)
+    return _tpc_factory
+
+
+def _verify_weights_quantizer_params(quant_method, weights_quantizer, params_shape, per_channel):
```
irenaby, line 107: The order of arguments is confusing, please put the tested object first. Also I usually add exp_ prefix to expected values to make it clear what is tested and what is the reference, just a suggestion.

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class TestPostTrainingQuantizationApi:
+    # TODO: add tests for:
+    #   1) activation only, W&A, LUT quantizer (separate)
+    #   2) extend to also test with different settings features (bc, snc, etc.)
+    #   3) advanced models and operators
+
+
+    def _verify_quantized_model_structure(self, model, q_model, quantization_info):
```
irenaby, line 129: Any reason this is inside the class and weights quantizer verification is outside? Anyway it's clearer to put the tests first

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+        assert q_model is not None and isinstance(q_model, keras.Model)
```
irenaby, line 130: Why specifically tests for None? isinstance covers it

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+        assert quantization_info is not None and isinstance(quantization_info, UserInformation)
+
+        # Assert quantized model structure
+        assert len([l for l in q_model.layers if isinstance(l, layers.BatchNormalization)]) == 0, \
+            "Expects BN folding in quantized model."
+        assert len([l for l in q_model.layers if isinstance(l, MetadataLayer)]) == 1, \
+            "Expects quantized model to have a metadata stored in a dedicated layer."
+        original_conv_layers = [l for l in model.layers if
```
irenaby, line 138: ensure this is not empty

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+                                isinstance(l, (layers.Conv2D, layers.DepthwiseConv2D, layers.Dense))]
+        quantized_conv_layers = [l for l in q_model.layers if isinstance(l, KerasQuantizationWrapper)]
```
irenaby, line 140: What about activations quantization? 
Also do you think stronger checks are needed, like correct layers are wrapped, model is connected correctly, etc?

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+
+    x = layers.Add()([x1, x2])
+
+    x = layers.Flatten()(x)
+    x = layers.Dense(units=10, activation='softmax')(x)
+
+    return keras.Model(inputs=inputs, outputs=x)
+
+
+def set_tpc(weights_quantizer, per_channel):
```
irenaby, line 71: Confusing arg, quantization method is not a quantizer

tests_pytest/integration_tests/keras/test_post_training_quantization.py
irenaby: Maybe check that the output model is valid (inference can be run)?

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+        quantized_conv_layers = [l for l in q_model.layers if isinstance(l, KerasQuantizationWrapper)]
+        assert len(original_conv_layers) == len(quantized_conv_layers), \
+            "Expects all conv layers from the original model to be wrapped with a KerasQuantizationWrapper."
+
+
+    @pytest.mark.parametrize("quant_method", [QuantizationMethod.POWER_OF_TWO,
+                                              QuantizationMethod.SYMMETRIC,
+                                              QuantizationMethod.UNIFORM])
+    @pytest.mark.parametrize("per_channel", [True, False])
+    @pytest.mark.parametrize("model", [model_basic(), model_residual()])
```
irenaby, line 149: model can be parametrized as fixture as well

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Assert quantization properties
+        quantized_conv_layers = [l for l in q_model.layers if isinstance(l, KerasQuantizationWrapper)]
+        for quantize_wrapper in quantized_conv_layers:
+            assert isinstance(quantize_wrapper.layer,
+                              (layers.Conv2D, layers.DepthwiseConv2D, layers.Dense, layers.Conv2DTranspose))
+
+            if isinstance(quantize_wrapper.layer, layers.DepthwiseConv2D):
+                weights_quantizer = quantize_wrapper.weights_quantizers[DEPTHWISE_KERNEL]
+                num_output_channels = (quantize_wrapper.layer.depthwise_kernel.shape[-1]
```
irenaby, line 166: You are inferring the expected value from the object you test, this is not a good practice. Also getting num channels from fw kernel shape doesn't ensure correct expected value. If both the test and the code access the wrong dimension, the test will pass. You defined the network, you know the correct values without extracting them from the model.

tests_pytest/integration_tests/keras/test_post_training_quantization.py
```diff
@@ -0,0 +1,178 @@
...
+        tpc = tpc_factory(quant_method, per_channel)
+        q_model, quantization_info = keras_post_training_quantization(model, rep_data_gen,
+                                                                      target_platform_capabilities=tpc)
+
+        self._verify_quantized_model_structure(model, q_model, quantization_info)
+
+        # Assert quantization properties
+        quantized_conv_layers = [l for l in q_model.layers if isinstance(l, KerasQuantizationWrapper)]
+        for quantize_wrapper in quantized_conv_layers:
+            assert isinstance(quantize_wrapper.layer,
```
irenaby, line 161: I think you can move this check to verify method. Also this is a pretty loose check. You can check that each original layer has a wrapper pointing to it.

tests_pytest/unit_tests/common/core/quantization/quantization_params_generation/test_symmetric_selection_activation.py
```diff
@@ -0,0 +1,77 @@
...
+
+err_methods_to_test = [e.name for e in QuantizationErrorMethod if e != QuantizationErrorMethod.HMSE]
+
+
+@pytest.mark.parametrize("error_method", err_methods_to_test)
+def test_symmetric_threshold_selection(error_method, hist):
+    counts, bins = hist
+
+    search_res = symmetric_selection_histogram(bins, counts, 2, 8, Mock(), Mock(), Mock(), Mock(),
+                                               MIN_THRESHOLD, QuantizationErrorMethod[error_method], True)
```
irenaby, line 59: If you pass the default value you cannot know if the passed value was used.
Also is_signed can accept 3 values, not only 2 that you test. You need to design the tests in a way that you can distinguish whether it was computed or taken from a flag.

tests_pytest/unit_tests/common/core/quantization/quantization_params_generation/test_symmetric_selection_activation.py
```diff
@@ -0,0 +1,77 @@
...
+    count, bins = np.histogram(x, bins=num_bins)
+
+    return count, bins
+
+
+err_methods_to_test = [e.name for e in QuantizationErrorMethod if e != QuantizationErrorMethod.HMSE]
+
+
+@pytest.mark.parametrize("error_method", err_methods_to_test)
+def test_symmetric_threshold_selection(error_method, hist):
```
irenaby, line 55: This is not really a unit test, it's more of a basic sanity. Let's talk.

tests_pytest/unit_tests/common/core/quantization/quantization_params_generation/test_symmetric_selection_activation.py
```diff
@@ -0,0 +1,77 @@
...
+    size = (32, 32, 3)
+    num_bins = 2048
+    x = np.random.uniform(-7, 7, size=size).flatten()
+    count, bins = np.histogram(x, bins=num_bins)
+
+    return count, bins
+
+
+@pytest.fixture
+def bounded_hist():
```
irenaby, line 39: What is meant by bounded and why is it interesting?

tests_pytest/unit_tests/common/core/quantization/quantization_params_generation/test_symmetric_selection_activation.py
```diff
@@ -0,0 +1,77 @@
...
+    np.random.seed(42)
+    size = (32, 32, 3)
+    num_bins = 2048
+    x = np.random.uniform(-7, 7, size=size).flatten()
+    e_x = np.exp(x - np.max(x))
```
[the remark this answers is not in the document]
ofirgo: I want an input that meets some conditions, but basically, I copied the same input from the old test to maintain the test functionality and coverage, and I don't recall what is the reason for these specific values.

tests_pytest/unit_tests/common/core/quantization/quantization_params_generation/test_symmetric_selection_activation.py
ofirgo: again, I copied the behavior from the old test, and I'm not quite sure what was the purpose.
Here I can guess that since the operator is softmax the feature map values are bounded by 1 so the symmetric threshold should be close to 1.

ofirgo:
@irenaby 
I removed the symmetric selection tests for now, just adding the ptq e2e tests in this PR (that's why I marked your comments as resolved).
Still need to fix the remaining comments so no need for reviewing yet.
One question in the meantime, how would you name the file of this kind of e2e tests? all e2e tests that want to check some ptq behavior under certain setting would call the "keras_post_training_quantization" API, but there are tens of tests like this, so we put all of them under the same file? or name the file according to the setting that we are testing (in this case something like "test_ptq_quantization_setting" since testing for different quantizers and qc options)

irenaby:
> @irenaby I removed the symmetric selection tests for now, just adding the ptq e2e tests in this PR (that's why I marked your comments as resolved). Still need to fix the remaining comments so no need for reviewing yet. One question in the meantime, how would you name the file of this kind of e2e tests? all e2e tests that want to check some ptq behavior under certain setting would call the "keras_post_training_quantization" API, but there are tens of tests like this, so we put all of them under the same file? or name the file according to the setting that we are testing (in this case something like "test_ptq_quantization_setting" since testing for different quantizers and qc options)

I would say definitely not under the same file. I think a file per some aspect that is tested, that can contain multiple tests that all relate to that aspect. Also I suggest we create a separate test directory for each api call (maybe except apis with a single test file like ru data). And maybe api tests is a better name than e2e, not that it really matters.

tests_pytest/integration_tests/keras/test_post_training_quantization.py
ofirgo: I'm not sure how what you suggesting is different (also, how to do it)?

PR #1385, opened 2025-03-13
MCT FX Tracer

yarden-yagil-sony:
## Pull Request Description:
Support models with custom layers


model_compression_toolkit/core/pytorch/reader/reader.py
```diff
@@ -13,19 +13,31 @@
...
-import torch
-from torch.fx import symbolic_trace
 from torch.fx.passes.shape_prop import ShapeProp
+from torch.fx import Tracer, GraphModule
 
 from model_compression_toolkit.logger import Logger
 from model_compression_toolkit.core.common import Graph
 from model_compression_toolkit.core.pytorch.reader.graph_builders import edges_builder, nodes_builder
 from model_compression_toolkit.core.pytorch.utils import set_model
+from sony_custom_layers.pytorch import CustomLayer
```
haihabi, line 26: Make sure that sony_custom_layers is part of MCT requirement

requirements.txt
```diff
@@ -10,4 +10,5 @@ matplotlib<3.10.0
 scipy
 protobuf
 mct-quantizers==1.5.2
-pydantic<2.0
\ No newline at end of file
+pydantic<2.0
+sony-custom-layers>0.2.0
```
haihabi, line 14: Why not exact minor ?

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2pytorch.py
```diff
@@ -97,7 +99,9 @@ def __init__(self):
             OperatorSetNames.L2NORM: [LayerFilterParams(torch.nn.functional.normalize,
                                                         Eq('p', 2) | Eq('p', None))],
             OperatorSetNames.SSD_POST_PROCESS: [],  # no such operator in pytorch
-            OperatorSetNames.COMBINED_NON_MAX_SUPPRESSION: []  # no such operator in pytorch
+            OperatorSetNames.COMBINED_NON_MAX_SUPPRESSION: [multiclass_nms, multiclass_nms_with_indices, MulticlassNMS,
```
haihabi, line 102: Why do we need multiclass_nms and multiclass_nms_with_indices if it is not support by fx trace ?

model_compression_toolkit/core/common/hessian/hessian_scores_calculator.py
```diff
@@ -85,6 +87,9 @@ def unfold_tensors_list(tensors_to_unfold: Any) -> List[Any]:
         """
         unfold_tensors = []
         for tensor in tensors_to_unfold:
+            if isinstance(tensor, NMSResults) or isinstance(tensor, NMSWithIndicesResults):
```
haihabi, line 90: Why not resolve this for a general tuple ?

haihabi (commented):
Few comment and questions

requirements.txt
ofirgo, line 14: Why is it a mandatory requirement?
Does sony-custom-layers automatically install any framework-dependent packages?

currently, MCT accesses sony-custom-layers only under the condition `FOUND_SONY_CUSTOM_LAYERS`, why not continue with this approach?

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2pytorch.py
```diff
@@ -97,7 +99,9 @@ def __init__(self):
...
+                                                            MulticlassNMSWithIndices],  # no such operator in pytorch
+            OperatorSetNames.BOX_DECODE: [FasterRCNNBoxDecode]
```
ofirgo, line 104: You need to add this key to `attach2keras` as well if you define it as part of the schema's built-in operator sets

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2pytorch.py
ofirgo, line 102: regarding my previous comment about the dependency in the requirement - you can see in this file that MCT accesses custom layers only under condition, I think that we should continue follow this approach here

model_compression_toolkit/core/common/hessian/hessian_scores_calculator.py
ofirgo: I agree, no need to check types or import custom layers here

model_compression_toolkit/core/pytorch/back2framework/pytorch_model_builder.py
```diff
@@ -138,8 +138,8 @@ def _run_operation(n: BaseNode,
                          for t in input_tensors]
         _tensor_input_allocs = None
 
-    if isinstance(n, FunctionalNode) and n.inputs_as_list:
-        if isinstance(op_func, PytorchQuantizationWrapper):
+    if n.inputs_as_list:
```
ofirgo, line 141: you changed the condition, there is an `else` statement, intended for the case where:
`not (isinstance(n, FunctionalNode) and n.inputs_as_list)`,
after the change, the "else" will be triggered for `not n.inputs_as_list`, which is not the same.
Why is this change needed? maybe it can be solved in another way

model_compression_toolkit/core/pytorch/reader/reader.py
ofirgo: Please see my comment regarding the requirements, I don't understand why to make it mandatory

model_compression_toolkit/core/pytorch/reader/reader.py
```diff
@@ -13,19 +13,31 @@
...
+
+    class MCTTracer(Tracer):
+        def is_leaf_module(self, m: torch.nn.Module, module_qualified_name: str) -> bool:
+            if isinstance(m, CustomLayer):
+                return True
+            return super().is_leaf_module(m, module_qualified_name)
+
+    tracer = MCTTracer()
+    graph = tracer.trace(model)
+    name = (model.__class__.__name__ if isinstance(model, torch.nn.Module) else model.__name__)
```
ofirgo, line 39: Can you explain the purpose of this line? seems like a hack, so I would prefer to solve it differently if we can

model_compression_toolkit/core/pytorch/reader/reader.py
```diff
@@ -87,7 +99,7 @@ def fx_graph_module_generation(pytorch_model: torch.nn.Module,
     set_model(pytorch_model)
 
     try:
-        symbolic_traced = symbolic_trace(pytorch_model)
+        symbolic_traced = _trace_model(pytorch_model)
```
ofirgo, line 102: Just a suggestion, but maybe we keep the previous behavior and just run the alternative tracing if it is necessary (meaning, if we have a model that uses custom layers)?
I wasn't convinced that this tracer is better suited to MCT's general operation

model_compression_toolkit/core/pytorch/reader/graph_builders.py
```diff
@@ -318,7 +318,7 @@ def nodes_builder(model: GraphModule,
                 Logger.critical(f'Found FX nodes in framework attributes of {node.name}. This node type should not contain any.')  # pragma: no cover
 
             graph_node_type = BaseNode
-            kwargs = {}
+            kwargs = {INPUTS_AS_LIST: inputs_as_list}
```
ofirgo, line 321: This shouldn't be the case in a non-functional node.
If this is specific to some custom layer that is not functional but expects input as list then it needs to be handled separately (or modified in custom layers, because it does not follow the protocol of torch modules)

requirements.txt
Idan-BenAmi: 1. The sony-custom-layers doesn't automatically install framework dependent package. 
2. the version is incorrect (should be 0.4.0 once publish)
3. I think that you're right about the condition. Yarden can you check if it's possible?

model_compression_toolkit/target_platform_capabilities/schema/v1.py
```diff
@@ -74,6 +74,7 @@ class OperatorSetNames(str, Enum):
     TOPK = "TopK"
     FAKE_QUANT = "FakeQuant"
     COMBINED_NON_MAX_SUPPRESSION = "CombinedNonMaxSuppression"
+    BOX_DECODE = "BoxDecode"
```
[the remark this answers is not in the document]
ofirgo: Only if it includes a 'BoxDecode' operator set. That will be the case every time that we decide to add an operator set name to the built-in set of names. I don't think that this was our original intention.

model_compression_toolkit/core/common/hessian/hessian_scores_calculator.py
yarden-yagil-sony: good suggestion,  updated to a general condition

requirements.txt
yarden-yagil-sony: because we check if a layer is a custom layer in the tracing phase
`model_compression_toolkit/core/pytorch/reader/reader.py:33`

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2pytorch.py
yarden-yagil-sony: added with empty list since we don't support this new custom- layer feature in keras, let me now if it's ok

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2pytorch.py
yarden-yagil-sony: see previous reply ;)

model_compression_toolkit/core/pytorch/reader/reader.py
yarden-yagil-sony: It extracts the name of a model, handling the possibility that the model might be a is a torch.nn.Module or just a function. 
 This line is taken from original torch symbolic_trace implementation so I prefer to keep it that way to avoid possible issues

model_compression_toolkit/core/pytorch/reader/reader.py
yarden-yagil-sony: The implementation is almost the same is torch implementation, I just added the custom layer check so everything should by ok. You can compare my `_trace_model`  implementation to [torch implementation](https://github.com/pytorch/pytorch/blob/e6afb51805d88547803252e4fb1a2757624d590c/torch/fx/_symbolic_trace.py#L1257)

model_compression_toolkit/core/pytorch/reader/reader.py
ofirgo: please add a comment to the code explaining this

model_compression_toolkit/core/pytorch/reader/reader.py
ofirgo: The key is in the word "almost" 😉 
I'm more concerned about future changes to this code, that would affect any user even if he doesn't use custom layers.
Also, that would eliminate future modifications to fx.symbolic_trace.
I think that supporting custom layers should be an addition, not a replacement, because this is a very specific case.

model_compression_toolkit/core/pytorch/back2framework/pytorch_model_builder.py
yarden-yagil-sony: You are correct!
I had a layer (custom layer) with inputs as list that is not a FunctionalNode/PytorchQuantizationWrapper, which led to error because it reached to the default behavior (the else->else) which is wrong for this case. 
I updated the condition, let me know what do think :)

model_compression_toolkit/core/pytorch/reader/reader.py
yarden-yagil-sony: I believe that after the discussion about adding custom layers as a requirement we will use the new fx tracing as the only option, deprecating the previous one. 
Let me know if you agree so I'll mark it as resolved

ofirgo: yes you can remove the old symbolic trace

model_compression_toolkit/core/pytorch/back2framework/pytorch_model_builder.py
yarden-yagil-sony: managed to create the custom layer without inputs as list, so deleted the changes in this file

requirements.txt
```diff
@@ -10,4 +10,5 @@ matplotlib<3.10.0
 scipy
 protobuf
 mct-quantizers==1.5.2
-pydantic<2.0
\ No newline at end of file
+pydantic<2.0
+sony-custom-layers==0.4.0
```
ofirgo, line 14: didn't we say it needs to be >=0.4.0?

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2pytorch.py
```diff
@@ -97,7 +98,8 @@ def __init__(self):
             OperatorSetNames.L2NORM: [LayerFilterParams(torch.nn.functional.normalize,
                                                         Eq('p', 2) | Eq('p', None))],
             OperatorSetNames.SSD_POST_PROCESS: [],  # no such operator in pytorch
-            OperatorSetNames.COMBINED_NON_MAX_SUPPRESSION: []  # no such operator in pytorch
+            OperatorSetNames.COMBINED_NON_MAX_SUPPRESSION: [MulticlassNMS, MulticlassNMSWithIndices],  # no such operator in pytorch
```
ofirgo, line 101: remove the comment from this line

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2keras.py
```diff
@@ -104,13 +103,7 @@ def __init__(self):
...
         }
 
-        if FOUND_SONY_CUSTOM_LAYERS:
-            self._opset2layer[OperatorSetNames.SSD_POST_PROCESS] = [SSDPostProcess]
-        else:
-            # If Custom layers is not installed then we don't want the user to fail, but just ignore custom layers
-            # in the initialized framework TPC
-            self._opset2layer[OperatorSetNames.SSD_POST_PROCESS] = []
-
+        self._opset2layer[OperatorSetNames.SSD_POST_PROCESS] = [SSDPostProcess]
```
ofirgo, line 106: you can add this to the dict initialization like all other operators, now that the `FOUND_SONY_CUSTOM_LAYERS` is removed

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2keras.py
```diff
@@ -93,6 +91,7 @@ def __init__(self):
             OperatorSetNames.TOPK: [tf.nn.top_k],
             OperatorSetNames.FAKE_QUANT: [tf.quantization.fake_quant_with_min_max_vars],
             OperatorSetNames.COMBINED_NON_MAX_SUPPRESSION: [tf.image.combined_non_max_suppression],
+            OperatorSetNames.BOX_DECODE: [],
```
ofirgo, line 94: add a comment like the other empty lists

model_compression_toolkit/target_platform_capabilities/schema/v1.py
ofirgo, line 77: Need to create Schema v2 instead of modifying v1.
I prefer it to be on a separate PR, please remove the schema change from this PR and open a new one with:
- creating a new 'v2' fuel under schema.
- copying and modifying `class OperatorSetNames` with the new operator set name.
- importing all other, unchanged classes from v1.
- setting `mct_current_schema` references to v2.

we'll merge it before this PR, so the attack2fw changes won't crash the nightly.

model_compression_toolkit/core/pytorch/utils.py
```diff
@@ -112,4 +112,13 @@ def clip_inf_values_float16(tensor: Tensor) -> Tensor:
     # Replace inf values with max float16 value
     tensor[inf_mask] = MAX_FLOAT16 * torch.sign(tensor[inf_mask])
 
-    return tensor
\ No newline at end of file
+    return tensor
+
+
+def is_tuple_of_tensors(obj):
```
ofirgo, line 118: Plase add function description and type hints

model_compression_toolkit/core/pytorch/reader/reader.py
```diff
@@ -13,19 +13,32 @@
...
+from torch.fx import Tracer, GraphModule
 
 from model_compression_toolkit.logger import Logger
 from model_compression_toolkit.core.common import Graph
 from model_compression_toolkit.core.pytorch.reader.graph_builders import edges_builder, nodes_builder
 from model_compression_toolkit.core.pytorch.utils import set_model
+from sony_custom_layers.pytorch import CustomLayer
+
+
+def _trace_model(model: torch.nn.Module) -> GraphModule:
```
ofirgo, line 29: Please add function description

ofirgo (changes requested):

model_compression_toolkit/target_platform_capabilities/schema/v1.py
yarden-yagil-sony: [Schema - V2 PR](https://github.com/sony/model_optimization/pull/1396)

model_compression_toolkit/target_platform_capabilities/targetplatform2framework/attach2pytorch.py
```diff
@@ -97,7 +98,8 @@ def __init__(self):
             OperatorSetNames.L2NORM: [LayerFilterParams(torch.nn.functional.normalize,
                                                         Eq('p', 2) | Eq('p', None))],
             OperatorSetNames.SSD_POST_PROCESS: [],  # no such operator in pytorch
-            OperatorSetNames.COMBINED_NON_MAX_SUPPRESSION: []  # no such operator in pytorch
+            OperatorSetNames.COMBINED_NON_MAX_SUPPRESSION: [MulticlassNMS, MulticlassNMSWithIndices],
+            OperatorSetNames.BOX_DECODE: [FasterRCNNBoxDecode]
```
ofirgo, line 102: This should be removed from this PR for now, right?

requirements.txt
ofirgo: I don't really care, I thought @Idan-BenAmi said that we should set >=.
I also think it is better because that way we won't need to update the requirement every CL release

ofirgo (approved):

PR #1387, opened 2025-03-16
Fix resource utilization data computation, remove requires_mix_precision check

irenaby:
## Pull Request Description:
- Compute resource utilization based on default quantization nbits, instead of hard-coded 8bit. For BOPS use default nbits as well (instead of float). 
- Remove requires_mixed_precision check. If target RU is passed, MP is enabled automatically (for 8 -> 16 bit activation) unless only weights are restricted. The check has an overhead of an additional run of graph preparation, and can be beneficial only if target weights utilization is passed and the model already satisfies it. Current implementation is also faulty (should use default bits instead of max bits). Considering all of the above it was decided to remove the check.
- Add unittest for compute_resource_utilization_data and API tests for resource utilization frameworks facades.


tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
```diff
@@ -0,0 +1,97 @@
...
+import pytest
+
+from model_compression_toolkit.core import CoreConfig, ResourceUtilization, QuantizationErrorMethod
+from model_compression_toolkit.target_platform_capabilities import AttributeQuantizationConfig, OpQuantizationConfig, \
+    Signedness, QuantizationConfigOptions, OperatorsSet, OperatorSetNames, TargetPlatformCapabilities, QuantizationMethod
+from model_compression_toolkit.target_platform_capabilities.constants import KERNEL_ATTR, BIAS_ATTR
+from tests_pytest._test_util.tpc_util import build_mp_config_options_for_kernel_bias_ops
+
+
+def get_tpc():
```
elad-c, line 29: Add comment describing the main characteristics of this TPC

tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+class BaseRUDataFacadeTest(abc.ABC):
+    bhwc_input_shape = (1, 8, 12, 8)
+
+    # api to test
+    fw_ru_data_facade: Callable
+
+    @abc.abstractmethod
+    def _build_model(self, input_shape, out_chan, kernel, const):
+        """ conv -> relu -> (add + const)
```
elad-c, line 68: Can reomve comment, or make it more general

tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+        input_shape = (1, 8, 12, 8)
+        model = self._build_model(input_shape=input_shape, out_chan=16, kernel=3, const=np.array([5]))
+
+        def repr_data_gen():
+            yield [np.random.random(input_shape)]
+
+        facade = self.__class__.fw_ru_data_facade
+        ru = facade(model, repr_data_gen, core_cfg, tpc)
+        assert ru == ResourceUtilization(activation_memory=(16*10*6*2)*8/8,
```
elad-c, line 92: This is very specific. will it work with any model? if not, why is _build_model not implemented here?

tests_pytest/common_tests/unit_tests/core/mixed_precision/resource_utilization_tools/test_resource_utilization_data.py
```diff
@@ -0,0 +1,75 @@
...
+    compute_resource_utilization_data
+
+
+class TestResourceUtilizationData:
+    @pytest.mark.parametrize('error_method', [QuantizationErrorMethod.MSE, QuantizationErrorMethod.HMSE])
+    def test_resource_utilization_data(self, fw_info_mock, fw_impl_mock, error_method, mocker):
+        core_cfg = CoreConfig()
+        core_cfg.quantization_config.weights_error_method = error_method
+        core_cfg.bit_width_config = BitWidthConfig([1, 2])
+        core_cfg_orig = copy.deepcopy(core_cfg)
```
elad-c, line 33: Should this happen before changing it?

tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
```diff
@@ -0,0 +1,97 @@
...
+    @pytest.mark.parametrize('error_method', [QuantizationErrorMethod.HMSE, QuantizationErrorMethod.MSE])
+    def test_resource_utilization_data_facade(self, error_method):
+        """ Integration test for resource data utilization user API.
+            We include HMSE since it has special handling (not allowed without gptq, so changed to MSE,
+            and make sure this change doesn't affect the original config. """
+        tpc = get_tpc()
+
+        core_cfg = CoreConfig()
+        core_cfg.quantization_config.weights_error_method = error_method
+        core_cfg_orig = copy.deepcopy(core_cfg)
```
elad-c, line 82: Shouldn't this happen before changing it?

tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
irenaby: This is a specific test testing a specific case, it's not a general test base class, that's why its under _fw_tests_common_base. Each concrete class implements the same specific model in torch/tf. While I'm not really thrilled about this, an alternative would be a full replication of the test most of which is identical (or defining a fw agnostic graph converted automatically to fw, which is even worse). This comment was supposed to make it clear, but apparently I failed :)

tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
irenaby: See comment above

tests_pytest/common_tests/unit_tests/core/mixed_precision/resource_utilization_tools/test_resource_utilization_data.py
irenaby: No, first I set up the config for the test, then I save its copy before it is used.

tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
irenaby: Added comments and documentation to the test. Hopefully it clarifies things.

tests_pytest/_fw_tests_common_base/base_ru_data_facade_test.py
irenaby: No

PR #1393, opened 2025-03-23
Fix virtual node construction with duplicate weights attributes

irenaby:
## Pull Request Description:
When creating the combined virtual node, weights from both activation and weights nodes are taken, as long as only the weights node contains kernel, so that weight utilization is computed correctly. If was assumed and validated that the activation node doesn't contain same weight attributes as the weights node, which is wrong, e.g. layer norm in torch contains weight and bias, same as conv.
This PR fixes this case. If activation and weight nodes contain duplicate weight attribute or index, a new unique name or index is created for the activation node and node's quantization cfg attribute and positional dicts are updated accordingly.

### DIFF
--- a/tests_pytest/_test_util/graph_builder_utils.py
+++ b/tests_pytest/_test_util/graph_builder_utils.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-from typing import Union, Iterable
+from typing import Union, Iterable, List
 
 from mct_quantizers import QuantizationMethod
 from model_compression_toolkit.core import QuantizationConfig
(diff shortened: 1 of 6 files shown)


ofirgo (approved):

tests_pytest/_test_util/graph_builder_utils.py
```diff
@@ -61,10 +69,24 @@ def full_attr_name(canonical_name: Union[str, dict, Iterable]):
     return canonical_name.__class__([convert(name) for name in canonical_name])
 
 
-def build_qc(a_nbits=8, a_enable=True, w_attr=None, pos_attr=(32, False, ())):
-    """ Build quantization config for tests.
-        w_attr contains {canonical name: (nbits, q_enabled)}
-        pos_attr: (nbits, q enabled, indices) """
+def build_nbits_qc(a_nbits=8, a_enable=True, w_attr=None, pos_attr=(32, False, ()),
```
irenaby, line 72: I'm keeping it here since it is used in several tests, but renamed and added description so it's clear(er) that this is not a generic qc builder

PR #1400, opened 2025-04-02
Replace the unitest version of test_tpc with a pytest version.

lior-dikstein:


ofirgo:
@lior-dikstein 
Does this PR include the new test we discussed that tries to run "load_tpc"  with a TPC of each of the existing schemas? (currently should only be with v1, when changing current schema to v2 we'll need to add it to the test as well)

PR #1403, opened 2025-04-06
Simplify MP manager - LP solver candidates representation, add MP, LP unit tests and MP e2e tests

irenaby:
## Pull Request Description:
Fix find_min/max_candidate graph's method.
Simplify data representation in LP solver. 
Convert configurations list in topological order to explicit dict from node to index where convenient.
Add unit tests for LP Solver
Add unit tests for MPSearchManager
Add E2E tests for mixed precision search, via ptq api, for torch.


model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
```diff
@@ -75,34 +77,44 @@ def __init__(self,
...
 
         self.mp_topo_configurable_nodes = self.mp_graph.get_configurable_sorted_nodes(fw_info)
-        self.layer_to_bitwidth_mapping = self.get_search_space()
 
         self.ru_targets = target_resource_utilization.get_restricted_targets()
         self.ru_helper = MixedPrecisionRUHelper(self.mp_graph, fw_info, fw_impl)
 
-        self.min_ru_config = self.mp_graph.get_min_candidates_config(fw_info)
-        self.max_ru_config = self.mp_graph.get_max_candidates_config(fw_info)
+        self.min_ru_config: Dict[BaseNode, int] = self.mp_graph.get_min_candidates_config(fw_info)
```
irenaby, line 84: No functional changes in this file. The main change is moving from a list to a dict which is more robust and easier to use.

model_compression_toolkit/core/common/mixed_precision/search_methods/linear_programming.py
```diff
@@ -30,23 +32,23 @@ class MixedPrecisionIntegerLPSolver:
             candidates_ru: resource utilization per candidate.
             ru_constraints: resource utilization constraints corresponding to 'candidates_ru'.
     """
-    def __init__(self, layer_to_sensitivity_mapping: Dict[int, Dict[int, float]],
+    def __init__(self,
```
irenaby, line 35: No functional change, only simplified representation - removed unnecessary inner dict whose keys always contain serial indices, and related code clean up

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -488,49 +488,35 @@ def get_total_output_params(self) -> float:
...
-        return [i for i, a_n_bits in min_candidates]
-
-    def find_max_candidates_indices(self) -> List[int]:
+        Returns:
+            The index of the minimal bit-width candidate.
         """
-        Returns a list with potential maximal candidates.
-        A potential maximal candidate is a candidate which its weights_n_bits and activation_n_bits pair is
-        on the Pareto Front, i.e., there is no other candidates that its n_bits pair is lower in both entries.
+        aw_nbits = [(c.activation_quantization_cfg.activation_n_bits,
```
ofirgo, line 496: Until now we assumed the ordering is: first - by weights bits, second - by activation bits.
doesn't this change this assumption (since you put activation first in the tuple)>

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
```diff
@@ -129,48 +141,42 @@ def _get_relative_ru_constraint_per_mem_element(self) -> Dict[RUTarget, np.ndarr
...
-
+            Mapping from nodes to their bitwidth candidates sensitivity.
         """
 
         Logger.info('Starting to evaluate metrics')
-        layer_to_metrics_mapping = {}
 
-        compute_metric = self.sensitivity_evaluator.compute_metric
+        def compute_metric(cfg, node_idx=None, baseline_cfg=None):
+            return self.sensitivity_evaluator.compute_metric(list(cfg.values()),
```
ofirgo, line 159: It's a bit difficult to track all the changes here, but again - does `cfg.values()` have any guarantee for the layers order? because `compute_metric` expects the values to match the topo order of the configurable nodes.
Even if it does, I think that we should use an explicit conversion from dict to cfg list every time we do this (to be on the safe side and, more important, to make this assumption explicit)

model_compression_toolkit/core/keras/back2framework/mixed_precision_model_builder.py
```diff
@@ -178,11 +174,7 @@ def mixed_precision_activation_holder(self, n: BaseNode) -> KerasActivationQuant
                 # if the node doesn't have a kernel attribute, we only sort by activation_n_bits.
                 n.sort_node_candidates(self.fw_info)
 
-                max_cfg_candidates = n.find_max_candidates_indices()
-                assert len(max_cfg_candidates) == 1, \
-                    f"A maximal config candidate must be defined, but some node have multiple potential maximal candidates"
-                max_candidate_idx = max_cfg_candidates[0]
-
+                max_candidate_idx = n.find_max_candidate_index()
```
ofirgo, line 177: I recall that we talked about removing some assumption, is this the one?
because now that I see it I'm not sure we can remove it.. do you have a case where this happens and doesn't cause issues?

ofirgo (changes requested):
I still need to go over the tests, but left some comments and questions for now

model_compression_toolkit/core/common/graph/base_node.py
irenaby: The tuple only serves for finding the max, we return the index of the candidate. Currently we assume (and validate) there is a candidate with both max weights and max activation, so the order doesn't matter. If we want to remove the assumption and define priority then the order in the tuple should be accordingly, but I think we agreed to keep the assumption.

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -22,6 +22,12 @@
 from model_compression_toolkit.defaultdict import DefaultDict
 
 
+# Default value to use for ops without kernel.
+# This is a weird default, but it's used all over the place, so for now only extract it to const so that it can be
+# referenced by variable instead of hard-coded.
+DEFAULT_KERNEL_ATTRIBUTES = [None]
```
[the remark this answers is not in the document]
irenaby: I don't like throwing all const into a single file. I think it makes more sense to define them where they belong logically.
But if you have a strong preference for a single file I won't argue. Let me know

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -493,6 +493,14 @@ def all_weight_attrs(self) -> List['WeightAttrT']:
         """
         return list(self.pos_attributes_config_mapping.keys()) + list(self.attributes_config_mapping.keys())
 
+    def get_all_weights_configs(self) -> Dict['WeightAttrT', AttributeQuantizationConfig]:
```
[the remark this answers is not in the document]
irenaby: I think we are not very consistent with names, but I use attr when i refer to weights attr as opposed to positional. Here it refers to all weights. It is used in finding max/min candidate - there is no assumption there that only kernels are quantized, all weights are taken into account

model_compression_toolkit/core/common/mixed_precision/search_methods/linear_programming.py
```diff
@@ -59,17 +61,14 @@ def run(self) -> List[int]:
...
 
         # Take the bitwidth index only if its corresponding indicator is one.
-        config = np.asarray(
-            [[nbits for nbits, indicator in nbits_to_indicator.items() if indicator.varValue == 1.0] for
-             nbits_to_indicator
-             in self.layer_to_indicator_vars_mapping.values()]
-        ).flatten()
-
-        return config.tolist()
+        mp_config = {
```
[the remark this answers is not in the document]
irenaby: core runner doesn't  interact with LP. MP facade returns a list as expected. I think dict is more robust and clearer, but I didn't want to make changes outside mp in this PR

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_facade.py
```diff
@@ -107,4 +107,4 @@ def search_bit_width(graph: Graph,
     if mp_config.refine_mp_solution:
         result_bit_cfg = greedy_solution_refinement_procedure(result_bit_cfg, search_manager, target_resource_utilization)
 
-    return result_bit_cfg
+    return list(result_bit_cfg.values())
```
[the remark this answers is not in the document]
irenaby: Per python specifications starting 3.7 dict is ordered by order of insertion.

model_compression_toolkit/core/keras/back2framework/mixed_precision_model_builder.py
irenaby: The only assumption removed was about the order between candidates. We still assume there is a candidate with max activation and max weights, and a candidate with min activation and min weight. It's validated inside n.find_max_candidate_index

model_compression_toolkit/core/common/framework_info.py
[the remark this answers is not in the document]
ofirgo: I don't like constant scattered around in the code, where anyone can (intentionally or mistakenly) import them to anywhere. When they are all in the same file/module it is simpler to maintain and verify no misuse has occurred.
But I don't have strong preference so you may keep it as you'd like.

model_compression_toolkit/core/common/quantization/node_quantization_config.py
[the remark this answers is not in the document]
ofirgo: I think that in most places in the code when we say "weights" we refer to weights in general (or kernel specifically, but this is just a leftover from the time we only quantized kernels), and when we refer to attributes, we usually say "weights attributes". I think it's clearer, "all weights config" sounds to me like all weights quantization candidates, but maybe this is because I'm too much into this part of the code

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_facade.py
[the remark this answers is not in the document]
ofirgo: I think that counting on insertaion order is too loose, let's reconstruct the config list from the dictionary manually

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
ofirgo: I agree that working with a dict is more explicit and clean and that the topo sort is not uniquely defined, but currently the infrastructure of this algorithm is based on configuration that is supposed to be ordered according to the topo sort.
Even if eventually we will change this, for now I think that we need a manual dict2list conversation mechanism.

model_compression_toolkit/core/common/quantization/node_quantization_config.py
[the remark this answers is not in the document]
irenaby: It can't be all candidates as the whole class is within a single candidate, but your suggestion is more consistent with other methods name in the class so I'll rename it

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
irenaby: changed

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_facade.py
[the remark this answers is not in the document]
irenaby: changed

tests/keras_tests/non_parallel_tests/test_lp_search_bitwidth.py
```diff
@@ -45,151 +42,6 @@
...
-                                   0: ResourceUtilization(bops=3)}}
-        mock_search_manager = MockMixedPrecisionSearchManager(layer_to_ru_mapping, {RUTarget.BOPS})
-
-        bit_cfg = self._execute(mock_search_manager, target_resource_utilization=target_resource_utilization)
-
-        self.assertTrue(len(bit_cfg) == 1)
-        self.assertTrue(bit_cfg[0] == 1)
-
-
 class TestSearchBitwidthConfiguration(unittest.TestCase):
```
ofirgo, line 45: Is it complicated to migrate this test as well and get rid of this test file?

ofirgo (approved):

PR #1404, opened 2025-04-06
Schema Compatibility

yarden-yagil-sony:
## Pull Request Description:
* Add schema compatibility mechanism - given a tpc of older schema version, converts it to the current schema version
* Add tests to validate new schemas are added properly

In case a new schema is added, the following steps are required (lets say we update schema v4 with new schema v5):
1. add `_schema_v4_to_v5` to` schema_compatibility.py` file
2. add `_schema_v4_to_v5` to `conversion_map`
3. add `schema_v5` to `ALL_SCHEMA_VERSIONS`

If the user doesn't perform one of the above steps, the new tests will fail with a description of what needs to be done.


model_compression_toolkit/target_platform_capabilities/schema/v1.py
```diff
@@ -647,7 +647,7 @@ class TargetPlatformCapabilities(BaseModel):
     name: Optional[str] = "default_tpc"
     is_simd_padding: bool = False
 
-    SCHEMA_VERSION: int = 1
+    SCHEMA_VERSION: ClassVar[int] = 1
```
ofirgo, line 650: Not sure if this is consistent regarding pydantic typing validation. might cause issues so perhaps it is better to not change typing of attributes in an existing schema

model_compression_toolkit/target_platform_capabilities/schema/v1.py
```diff
@@ -697,4 +697,5 @@ def show(self):
         """
         Display the TargetPlatformCapabilities.
         """
-        pprint.pprint(self.get_info(), sort_dicts=False)
\ No newline at end of file
+        pprint.pprint(self.get_info(), sort_dicts=False)
+
```
ofirgo, line 701: remove new line (just because it is a shema file, it is better to not have unnecessary "messy" history, otherwise I wouldn't care)

model_compression_toolkit/target_platform_capabilities/schema/schema_compatability.py
```diff
@@ -0,0 +1,68 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+from typing import Any
+
+import model_compression_toolkit.target_platform_capabilities.schema.v1 as schema_v1
+import model_compression_toolkit.target_platform_capabilities.schema.v2 as schema_v2
+import model_compression_toolkit.target_platform_capabilities.schema.mct_current_schema as schema
+
```
ofirgo, line 20: add a comment saying this needs to be updated with all active schema versions

model_compression_toolkit/target_platform_capabilities/schema/schema_compatability.py
```diff
@@ -0,0 +1,68 @@
...
+    return schema.TargetPlatformCapabilities(default_qco=tpc.default_qco,
+                                             operator_set=tpc.operator_set,
+                                             fusing_patterns=tpc.fusing_patterns,
+                                             tpc_minor_version=tpc.tpc_minor_version,
+                                             tpc_patch_version=tpc.tpc_patch_version,
+                                             tpc_platform_type=tpc.tpc_platform_type,
+                                             add_metadata=tpc.add_metadata)
+
+
+def tpc_to_current_schema_version(tpc: schema.TargetPlatformCapabilities):
```
ofirgo, line 49: the type is not necessarily of `schema.TargetPlatformCapabilities` right? can be any schema version `TargetPlatformCapabilities`

model_compression_toolkit/target_platform_capabilities/schema/schema_compatability.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Given tpc instance of some schema version, convert it to the current MCT schema version.
+
+    In case a new schema is added to MCT, need to add a conversion function from the previous version to the new
+    version, e.g. if the current schema version was updated from v4 to v5, need to add _schema_v4_to_v5 function to
+    this file, than and add it to the conversion_map.
```
ofirgo, line 55: remove 'than' (or "then" and remove "and"?)

model_compression_toolkit/target_platform_capabilities/tpc_io_handler.py
```diff
@@ -15,12 +15,34 @@
...
+
+    try:
+        return schema.TargetPlatformCapabilities.parse_raw(data)
+    except ValueError as e:
+        raise ValueError(f"Invalid JSON for loading TargetPlatformCapabilities in '{tpc_path}': {e}.") from e
+    except Exception as e:
+        raise ValueError(f"Unexpected error while initializing TargetPlatformCapabilities: {e}.") from e
+
+
+def load_target_platform_capabilities(tpc_obj_or_path: Union[schema.TargetPlatformCapabilities, str]) -> schema.TargetPlatformCapabilities:
```
ofirgo, line 45: type of `tpc_obj_or_path` is not necessarily `schema.TargetPlatformCapabilities` since this is before possible conversion

ofirgo (changes requested):

model_compression_toolkit/target_platform_capabilities/schema/v1.py
yarden-yagil-sony: Without this change we can't access `SCHEMA_VERSION`, this change convert it to be like a static variable that can be accessed without initialized instance

ofirgo: Are you sure? It's not what I'm familiar with and also doesn't match what I find on the internet.

Anyway, this change can be problematic since, as far as I understand from reading, when you set a ClassVar in a pydantic class it is not serializable, but we need it to be.
In addition, if the converter are reading and using this attribute, and counting on it being an int, then this change will break previous converter versions that should support schema 1.

So, if using simple int is not working, then we need to figure out a different solution.

yarden-yagil-sony: Yes, here is a source explaining it:
https://stackoverflow.com/questions/70852331/how-to-define-class-attributes-after-inheriting-pydantics-basemodel
 
 However, it is correct that the ClassVar is not serializable. I removed the ClassVar and updated the code to use type instead of version number. I don't like it, especially because of the while loop, but it works. Let me know what do you think

model_compression_toolkit/target_platform_capabilities/tpc_io_handler.py
```diff
@@ -15,12 +15,34 @@
...
-from model_compression_toolkit.logger import Logger
-from model_compression_toolkit.target_platform_capabilities.schema.mct_current_schema import TargetPlatformCapabilities
-import json
+import model_compression_toolkit.target_platform_capabilities.schema.mct_current_schema as schema
+from model_compression_toolkit.target_platform_capabilities.schema.schema_compatability import is_tpc_instance, \
+    tpc_to_current_schema_version, tpc_or_str_type
 
 
-def load_target_platform_capabilities(tpc_obj_or_path: Union[TargetPlatformCapabilities, str]) -> TargetPlatformCapabilities:
+def _get_tpc_from_json(tpc_path):
```
lior-dikstein, line 23: need to add documentation to function

tests_pytest/common_tests/unit_tests/target_platform_capabilities/test_tpc.py
```diff
@@ -88,6 +108,17 @@ def test_valid_model_object(self, tpc):
         result = load_target_platform_capabilities(tpc)
         assert result == tpc
 
+    def test_new_schema(self, tpc):
+        """Tests that current schema is in all schemas list. This test validates new schema was added properly."""
+        assert type(tpc) in all_tpc_types, "Current schema need to be added to ALL_SCHEMA_VERSIONS"
```
lior-dikstein, line 113: This is the relevant code under test:
```
ALL_SCHEMA_VERSIONS = [schema_v1]  # needs to be updated with all active schema versions
all_tpc_types = tuple([s.TargetPlatformCapabilities for s in ALL_SCHEMA_VERSIONS])
```
I’m not sure this achieves the intended behavior. It seems like something might be missing, but perhaps I’m misunderstanding the context. Let’s discuss

model_compression_toolkit/target_platform_capabilities/schema/schema_compatability.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Checks if the given object is an instance of a TargetPlatformCapabilities
+    :param tpc_obj_or_path: Object to check its type
+    :return: True if the given object is an instance of a TargetPlatformCapabilities, False otherwise
+    """
+    return type(tpc_obj_or_path) in all_tpc_types
+
+
+def _schema_v1_to_v2(tpc: schema_v1.TargetPlatformCapabilities) -> schema_v2.TargetPlatformCapabilities:
+    """
```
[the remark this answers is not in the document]
lior-dikstein: The current conversion is simple and short, but keep in mind that it could become significantly more complex as more schemas are introduced.

Conceptually, this seems like something that defines a TPC based on a previous schema, which is why I believe it belongs in the schema itself.

That said, if placing it in the schema file isn't ideal, perhaps each schema could have its own package folder to better organize version-specific logic?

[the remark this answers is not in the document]
ofirgo: not sure that I understand what you're suggesting.
I think that it is a good idea to create a more robust mechanism for handling this conversion between schema versions.
However, this can be taken care of in a future PR after we decide on a mechanism (as long as it does not involve changing existing schema).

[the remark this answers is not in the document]
yarden-yagil-sony: Regarding implementing `tpc_to_current_schema_version` function in the schema itself, it was like that ([older_commit](https://github.com/sony/model_optimization/pull/1404/commits)), but after discussion we decided to move it out to avoid having logic inside the schema itself. Let me know if you prefer it the way it was.

Regarding splitting it to seperate packages, I agree that in the long term it's better. Are you both ok with the following structure?
```
--target_platform_capabiilities
    --schema
        mct_current_schema.py
        schema_functions.py
        schema_compatability.py
        --schema_v1
            v1.py
            v1_compatability.py (_schema_v1_to_v2 function inside )
        --schema_v2
            v2.py
            v2_compatability.py (at the moment empty)
```
Or do you want me it to keep it the way it is and change it in the future if needed?

@ofirgo @lior-dikstein

lior-dikstein (approved):

PR #1420, opened 2025-04-22
add support for custom metric function for mixed precision

itai-berman:
## Pull Request Description:
Add support for custom metric function for mixed precision. 
The function will  get the model_mp as input and return a metric score. 
model_mp will return all the model's outputs and not the interest points.


model_compression_toolkit/core/common/mixed_precision/mixed_precision_quantization_config.py
```diff
@@ -39,6 +39,7 @@ class MixedPrecisionQuantizationConfig:
 
     compute_distance_fn: Optional[Callable] = None
     distance_weighting_method: MpDistanceWeighting = MpDistanceWeighting.AVG
+    custom_metric_fn: Optional[Callable] = None
```
irenaby, line 42: Please add description to docstring, including the expected api of the function (args it accepts and what it should return).

model_compression_toolkit/core/common/mixed_precision/mixed_precision_search_manager.py
```diff
@@ -162,9 +162,15 @@ def topo_cfg(cfg: dict) -> list:
             return topo_cfg
 
         def compute_metric(cfg, node_idx=None, baseline_cfg=None):
-            return self.sensitivity_evaluator.compute_metric(topo_cfg(cfg),
-                                                             node_idx,
-                                                             topo_cfg(baseline_cfg) if baseline_cfg else None)
+            if self.sensitivity_evaluator.quant_config.custom_metric_fn is None:
```
irenaby, line 165: I think it's better to keep a single entry point to sensitivity evaluator and keep mp manager agnostic to this, i.e. call self.sensitivity_evaluator.compute_metric and let it decide what to do. No reason to spread the logic between two places.

model_compression_toolkit/core/common/mixed_precision/sensitivity_evaluation.py
```diff
@@ -187,6 +193,38 @@ def compute_metric(self,
         return self._compute_mp_distance_measure(ipts_distances, out_pts_distances,
                                                  self.quant_config.distance_weighting_method)
 
+    def compute_custom_metric(self,
```
irenaby, line 196: Consider uniting the two methods. configure -> compute default or custom metric -> configure back, instead of replicating configuration in two places. You can move line 193 before 189, so there shouldn't be a problem.

irenaby (approved):

tests_pytest/pytorch_tests/unit_tests/core/mixed_precision/test_custom_metric_function.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+@pytest.fixture
+def mock_model_configuration():
+    return Mock()
+
+
+@pytest.fixture
+def sensitivity_evaluator_factory():
+    def _create_sensitivity_evaluator(custom_metric_fn):
```
irenaby, line 49: What's the benefit of having this as a fixture? It complicates the definition, you need to pass it to test as a fixture, and then you call it anyway inside the test. You could just define it directly (or combine with get_sensitivity_evaluator) and call from test as a regular function/method.

irenaby (changes requested):

irenaby (approved):

PR #1422, opened 2025-04-22
Node Supported Bitwidth Filtering Bugfix

yarden-yagil-sony:
## Pull Request Description:
Filter out nodes without enable quantization and not preserving

### DIFF
--- a/model_compression_toolkit/core/common/quantization/set_node_quantization_config.py
+++ b/model_compression_toolkit/core/common/quantization/set_node_quantization_config.py
@@ -119,11 +119,16 @@ def filter_node_qco_by_graph(node: BaseNode,
             _next_nodes.extend(graph.get_next_nodes(n))
         next_nodes.append(n)
 
-    if len(next_nodes):
-        next_nodes_qc_options = [_node.get_qco(fqc) for _node in next_nodes]
-        next_nodes_supported_input_bitwidth = min([max_input_activation_n_bits(op_cfg)
+    if len(next_nodes) == 0:
+        return _base_config, _node_qc_options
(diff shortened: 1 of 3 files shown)


ofirgo:
Can you please explain what was the bug here?
@yarden-yagil-sony

model_compression_toolkit/core/common/quantization/set_node_quantization_config.py
```diff
@@ -119,11 +119,16 @@ def filter_node_qco_by_graph(node: BaseNode,
...
-        next_nodes_qc_options = [_node.get_qco(fqc) for _node in next_nodes]
-        next_nodes_supported_input_bitwidth = min([max_input_activation_n_bits(op_cfg)
+    if len(next_nodes) == 0:
+        return _base_config, _node_qc_options
+    next_nodes_qc_options = [_node.get_qco(fqc) for _node in next_nodes]
+    all_next_nodes_supported_input_bitwidth = [max_input_activation_n_bits(op_cfg)
                                                    for qc_opts in next_nodes_qc_options
-                                                   for op_cfg in qc_opts.quantization_configurations])
+                                                   for op_cfg in qc_opts.quantization_configurations
+                                               if op_cfg.enable_activation_quantization or op_cfg.quantization_preserving
```
yarden-yagil-sony, line 128: here the "if" condition was missing, which led to considering cq of nodes with  enable_activation_quantization=False and quantization_preserving=False. 
It leds to error when these irrelevant nodes support only 8 and not both 8 and 16 bit, resulting in a node manually set to bithwidth16 but support only 8 bit (since we take the minimum of preserving next nodes supported bitwidth, line 131).

PR #1435, opened 2025-05-05
Fix issue with computing max cut when SNC is used

reuvenperetz:
## Pull Request Description:
When running MCT with Shift Negative Activation Correction (SNC), affected layers automatically had their activation quantization bitwidth set to 16-bit, regardless of the configuration defined in the TPC.
But not reflected in the resource utilization data used for optimization.
This mismatch led to inaccurate memory estimation, potentially resulting in suboptimal bitwidth assignment during mixed-precision search.
This PR fixes this by adding the fusing patterns to the fusing info during the SNC substitution, updating the fusing info due to the SNC changes, so that during the resources computation it treats the SNC fusing patterns as fused op and ignores the intermediate tensorts inside the fused operator.


model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
```diff
@@ -45,6 +48,63 @@
 """
 
 
+def get_snc_tpc():
```
elad-c, line 51: remove tpc

model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
```diff
@@ -404,8 +467,26 @@ def shift_negative_function(graph: Graph,
 
         candidate_qc.activation_quantization_cfg.set_activation_quantization_param({THRESHOLD: activation_threshold,
                                                                                     SIGNED: False})
+
+        # TODO: do we ignore the TPC configuration about 'add' ops?
```
elad-c, line 471: maybe add an assert that verifies that ADD supports the bitwidth in the TPC

tests_pytest/_fw_tests_common_base/base_ru_integration_test.py
```diff
@@ -85,6 +88,44 @@ def build_tpc():
     return tpc, linear_w_min_nbit, linear_a_min_nbit, default_w_nbit, default_a_nbit, binary_out_a_bit
 
 
+def build_snc_tpc():
```
elad-c, line 91: Need to consider how to test this feature, through a fusing pattern or using a default TPC, because current TPCs don't include this fusing.

elad-c (changes requested):

tests_pytest/_fw_tests_common_base/base_ru_integration_test.py
reuvenperetz: The next PR contains e2e tests that will include this scenario.

model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
```diff
@@ -456,6 +399,8 @@ def shift_negative_function(graph: Graph,
                     graph.shift_stats_collector(bypass_node, np.array(shift_value))
 
     add_node_qco = add_node.get_qco(graph.fqc).quantization_configurations
+    assert original_non_linear_activation_nbits in [c.activation_n_bits for c in add_node_qco]
```
elad-c, line 402: Actually, maybe it's better to raise a ValueError with message, because a user can cause this error with the TPC he supplies.

model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
```diff
@@ -476,16 +421,15 @@ def shift_negative_function(graph: Graph,
 
     if len(previous_nodes) == 1:
         fused_candidates = [previous_nodes[0], non_linear_node, add_node]
+        graph.fusing_info.fusing_patterns.append([n.type for n in fused_candidates])
```
elad-c, line 424: is this needed EVERY SNC call? isn't it the same pattern most of the time?
maybe `fusing_patterns` should be a set

model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
```diff
@@ -476,16 +421,15 @@ def shift_negative_function(graph: Graph,
...
 
-        if graph.fusing_info.is_nodes_eligible_to_be_fused(fused_candidates):
-            fused_op_id = graph.fusing_info.generate_fused_op_id(fused_candidates)
-
-            # If the non-linear is already part of a fused op - remove the existing one before adding the new one
-            if graph.fusing_info.is_node_in_fused_op(non_linear_node):
-                existing_fused_op = graph.fusing_info.get_fused_node_name(non_linear_node.name)
-                graph.fusing_info.remove_fused_operation(existing_fused_op)
+        # If the non-linear is already part of a fused op - remove the existing one before adding the new one
+        if graph.fusing_info.is_node_in_fused_op(non_linear_node):
```
elad-c, line 428: what's the scenario for this?

reuvenperetz: If for example there is a fusion for 'conv->swish', but adding the 'add' should replace it to a new fusion of 'conv->swish->add'

model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
reuvenperetz: The graph can contain different patterns (if, for example, it has conv->swish and dw->prelu, different patterns are needed).

elad-c: of course, but in most cases,1 or 2 patterns will exist in a model

model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
```diff
@@ -399,7 +399,12 @@ def shift_negative_function(graph: Graph,
                     graph.shift_stats_collector(bypass_node, np.array(shift_value))
 
     add_node_qco = add_node.get_qco(graph.fqc).quantization_configurations
-    assert original_non_linear_activation_nbits in [c.activation_n_bits for c in add_node_qco]
+    add_supported_bitwidths = [c.activation_n_bits for c in add_node_qco]
+    if original_non_linear_activation_nbits not in add_supported_bitwidths:
```
elad-c, line 403: unittest?

elad-c (approved):

model_compression_toolkit/core/common/substitutions/shift_negative_activation.py
```diff
@@ -187,6 +187,65 @@ def remove_node_between_two_nodes(graph: Graph,
...
+            fused_candidates.extend(fusing_info.get_fused_nodes(fused_id))
+    fused_candidates.append(add_node)
+
+    # Remove duplicates while preserving order
+    fused_candidates = list(dict.fromkeys(fused_candidates))
+
+    # Remove existing fused ops involving prev_node or non_linear_node
+    for node in (prev_node, non_linear_node):
+        if fusing_info.is_node_in_fused_op(node):
+            fusing_info.remove_fused_operation(fusing_info.get_fused_node_name(node.name))
```
elad-c, line 241: Are op_id and fused_node_name the same thing? if so, it's a little confusing.

elad-c (approved):

PR #1438, opened 2025-05-07
Add an option to dynamically disable configurable quantizers, separate configuration api for activation and weights quantizers

irenaby:
## Pull Request Description:
* When setting bitwidth to configurable quantizer, None can now be passed to disable quantization.
* Setting bitwidth api was split to separate apis for activation and weights
* Remove SensotivityEvaluation factory from fw_impl, keep only the relevant quant classes.


model_compression_toolkit/core/keras/mixed_precision/configurable_activation_quantizer.py
```diff
@@ -96,6 +96,8 @@ def __call__(self,
         Returns:
             Quantized activation tensor.
         """
+        if self.active_quantization_config_index is None:
+            return inputs.numpy()
```
reuvenperetz, line 100: Is this supposed to be numpy? Why not tf tensor?

reuvenperetz (approved):

model_compression_toolkit/core/keras/mixed_precision/configurable_activation_quantizer.py
irenaby: I think the actual quantization methods use numpy. But I just complied with the existing interface, it's even has numpy as a return type hint.

PR #1447, opened 2025-05-23
Enable activation status collection on FLN node.

gouda-youichi:
## Pull Request Description:
Enable act status collect on fln node


tests_pytest/common_tests/unit_tests/test_model_collector.py
```diff
@@ -55,6 +55,8 @@ def test_create_stats_collector_for_node_activation_enabled(self, fw_info_mock):
         # Set up a fake node with activation quantization enabled and prior_info attributes.
         node = Mock()
         node.is_activation_quantization_enabled.return_value = True
+        node.is_quantization_preserving.return_value = False
```
ofirgo, line 58: why do you need the `is_quantization_preserving` mock here? where is it being checked during the stat collector execution? (maybe I missed it)

tests_pytest/common_tests/unit_tests/test_model_collector.py
```diff
@@ -68,13 +70,33 @@ def test_create_stats_collector_for_node_activation_disabled(self, fw_info_mock)
...
+        The test for FLN activation disabled is already covered by 
+            test_create_stats_collector_for_node_activation_disabled, so it is not checked here.
+        """
+        # Set up a fake node with activation quantization enabled and prior_info attributes.
+        node = Mock()
+        node.is_activation_quantization_enabled.return_value = False
+        node.is_quantization_preserving.return_value = False
+        node.is_fln_quantization.return_value = True
+        node.type = DummyLayer
+        node.prior_info = Mock(min_output=-8, max_output=9)
```
ofirgo, line 95: this is a good use case that you are testing with prior_info. please also add a test scenario where the prior_info is None

tests_pytest/common_tests/unit_tests/test_model_collector.py
```diff
@@ -272,3 +352,113 @@ def test_update_statistics_called(self, fw_impl_mock, fw_info_mock):
...
+        self.qc.activation_error_method = QuantizationErrorMethod.MSE
+        mc = ModelCollector(self.graph, fw_impl_mock, fw_info_mock, qc=self.qc, hessian_info_service=self.fake_hessian_info_service)
+
+        mc.infer(self.infer_input)
+        # Check that inference runs without requiring gradients.
+        fw_impl_mock.run_model_inference.assert_called_once_with(mc.model, self.infer_input, requires_grad=False)
+        # Confirm that the Hessian service is not used.
+        mc.hessian_service.fetch_hessian.assert_not_called()
+
+    def test_infer_with_hessian_in_FLN(self, fw_impl_mock, fw_info_mock):
```
ofirgo, line 430: There is no need to check FLN with hessian for stats collection.
In fact, I think we should forbid this in the code, and make FLN statistics to never use hessians for statistic collection (I'll add a comment about it separately).
So - you can modify the tests to remove the fake Hessain Service from the setup, and keep only the test without hessians.

model_compression_toolkit/core/common/model_collector.py
```diff
@@ -44,7 +44,7 @@ def create_stats_collector_for_node(node: common.BaseNode,
         Statistics collector for statistics collection for the node.
     """
 
-    if node.is_activation_quantization_enabled():
+    if node.is_activation_quantization_enabled() or node.is_fln_quantization():
```
ofirgo, line 47: To force FLN quant nodes to not run use hessians in statistic collection, I would suggest passing this information here to a new argument in `StatsCollector` and checking later where setting `compute_hessians` during `infer`.

ofirgo (changes requested):
Some small modifications required
(I accedentally closed and reopened the PR, ignore it)

tests_pytest/common_tests/unit_tests/test_model_collector.py
gouda-youichi: As you said, this is not necessary.
It is written here to explicitly state that this node is not preserving.
Should I delete it?

ofirgo: Yes. If it is not neccessary for the test to work or is not something that the test is validating then you may delete it (here and in the other tests as well)

model_compression_toolkit/core/common/model_collector.py
gouda-youichi: As we have discussed, this topic is postponed.

PR #1462, opened 2025-06-05
Fix check in model collector for quantized FLN node

ofirgo:
## Pull Request Description:
To decide whether to collect statistics for a node inside FLN, it is not enough to check its quantization mode, but also needs to verify if it is supposed to be quantized based on the noe config in the fusing information.

### DIFF
--- a/model_compression_toolkit/core/common/fusion/fusing_info.py
+++ b/model_compression_toolkit/core/common/fusion/fusing_info.py
@@ -202,6 +202,8 @@ class FusingInfo:
         """
         Returns a list of the nodes that are part but not the last node of an FLN.
         """
+        # TODO: the order of the nodes is not gurenteed when returned as dict from get_all_fused_operations -
+        #  then, removing the last one can cause issues
         return [node for nodes in self.get_all_fused_operations().values() for node in nodes[:-1]]
 
     def get_fused_op_quantization_config(self, op_id: str) -> OpQuantizationConfig:


model_compression_toolkit/core/common/fusion/fusing_info.py
```diff
@@ -198,9 +198,9 @@ def get_fused_nodes(self, op_id: str) -> Optional[List['BaseNode']]:
         """
         return self.fusing_data.get(op_id)
 
-    def get_nodes_to_disable_activation_quantization(self) -> List['BaseNode']:
+    def get_inner_fln_nodes(self) -> List['BaseNode']:
         """
-        Returns a list of the nodes that their activation quantization is disabled due to fusing.
+        Returns a list of the nodes that are part but not the last node of an FLN.
```
elad-c, line 203: seems that get_all_fused_operations returns a dict, which might not be ordered, and the we count on the order to remove the last node. isn't it a vulnerability?

model_compression_toolkit/core/common/fusion/fusing_info.py
```diff
@@ -228,6 +228,22 @@ def is_node_in_fused_op(self, node: 'BaseNode') -> bool:
...
+        Check whether a node inside an FLN and should be quantized.
+
+        Args:
+            node (BaseNode): The node to check.
+
+        Returns:
+            bool: True if the node is in any fused operation and should be quantized.
+        """
+        if self.is_node_in_fused_op(node):
+            node_q_cfg = self.fused_op_id_to_quant_config[self.node_to_fused_node_map[node.name]]
```
elad-c: isn't `node_name_to_fused_op_id` a better name than `node_to_fused_node_map`?

model_compression_toolkit/core/common/fusion/fusing_info.py
```diff
@@ -228,6 +228,22 @@ def is_node_in_fused_op(self, node: 'BaseNode') -> bool:
...
+            return node_q_cfg is not None and node_q_cfg.enable_activation_quantization
```
elad-c, line 245: must we carry the op_cfg, and not put the data on the node? probably require another quant_mode: FLN_NOQUANT...
It will make the changes in the model_collector redundant

model_compression_toolkit/core/common/fusion/fusing_info.py
ofirgo: I think you're right.
Anyway, this is not related to this PR, I'm adding a TODO and I'll talk with @reuvenperetz about it to make sure.

elad-c (approved):

PR #1464, opened 2025-06-08
Fix bug in export using FQ ONNX in replacing activation holder with more than 8 bits with identity

lapid92:

### DIFF
--- a/model_compression_toolkit/exporter/model_exporter/pytorch/fakely_quant_onnx_pytorch_exporter.py
+++ b/model_compression_toolkit/exporter/model_exporter/pytorch/fakely_quant_onnx_pytorch_exporter.py
@@ -78,10 +78,17 @@ if FOUND_ONNX:
             act_holder_list = [n for n, m in self.model.named_modules()
                                if isinstance(m, PytorchActivationQuantizationHolder) and
                                m.activation_holder_quantizer.num_bits > 8]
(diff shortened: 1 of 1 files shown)


model_compression_toolkit/exporter/model_exporter/pytorch/fakely_quant_onnx_pytorch_exporter.py
```diff
@@ -78,10 +78,17 @@ def export(self, output_names=None) -> None:
...
+            for act_holder in act_holder_list: # pragma: no cover
+                obj = self.model
+                attrs = act_holder.split(".")
+                for a in attrs[:-1]:
+                    obj = getattr(obj, a)
+                if hasattr(obj, attrs[-1]):
+                    delattr(obj, attrs[-1])
+                    setattr(obj, attrs[-1], torch.nn.Identity())
+                else:
+                    Logger.info(f"Failed to delete activation quantization holder {act_holder}, "
```
elad-c: Make the message more informative. Something like: "During removal of activation quantization of quantizer with more than 8 bits, deleting act blala failed, ...."

ofirgo: You don't verify that it is more than 8 bits, do you?
Do we want to delete it only if it has more than 8bit?

PR #1465, opened 2025-06-11
Framework info refactor

elad-c:
## Pull Request Description:
This PR refactors the fw_info throughout the code to a global class. That class is selected per framework i nthe facade, and is saved in a global "pointer" in common code. This pointer is set & retreived with common functions `get_fw_info` and `set_fw_info`.

### DIFF
--- a/model_compression_toolkit/core/common/graph/base_node.py
+++ b/model_compression_toolkit/core/common/graph/base_node.py
@@ -99,11 +99,11 @@ class BaseNode:
         self.prior_info = None
         self.has_activation = has_activation
         self.is_custom = is_custom
-        self._set_fw_node_attrs(layer_class, framework_attr)
+        self.node_fw_info = self._set_fw_node_attrs(layer_class, framework_attr)
 
     def _set_fw_node_attrs(self, node_type, framework_attr):
         fw_info = get_fw_info()
-        self.node_fw_info = None if fw_info is None else NodeFrameworkInfo(
+        return None if fw_info is None else NodeFrameworkInfo(
             fw_info.get_kernel_channels(node_type),
             fw_info.get_out_channel_axis(node_type),
             fw_info.get_layer_min_max(node_type, framework_attr),


model_compression_toolkit/ptq/pytorch/quantization_facade.py
```diff
@@ -102,7 +103,7 @@ def pytorch_post_training_quantization(in_module: Module,
         if core_config.debug_config.bypass:
             return in_module, None
 
-        fw_info = DEFAULT_PYTORCH_INFO
+        set_fw_info(PyTorchInfo)
```
ofirgo: Why explicitly and not via a decorator like Irena suggested? seems cleaner, and we want to have as short and simple facades as possible.
If this is the plan for a separate PR then ignore me

model_compression_toolkit/ptq/pytorch/quantization_facade.py
```diff
@@ -34,7 +34,8 @@
 from model_compression_toolkit.metadata import create_model_metadata
 
 if FOUND_TORCH:
-    from model_compression_toolkit.core.pytorch.default_framework_info import DEFAULT_PYTORCH_INFO
+    from model_compression_toolkit.core.common.framework_info import set_fw_info
```
ofirgo: this import can be placed outside the "FOUND_TORCH" scope

model_compression_toolkit/core/pytorch/default_framework_info.py
```diff
@@ -26,73 +26,68 @@
...
+                                          (None, None))
 
+    """
+    Map a layer to its output channel axis.
+    Where axis=-1 is the last axis
+    """
+    out_channel_axis_mapping = DefaultDict({Conv2d: 1,
+                                            Linear: -1,
+                                            ConvTranspose2d: 1},
+                                           1)
```
ofirgo: I'm assuming taking care of the default value and the functions mapping below is left for a separate PR?
If so, please add all these open items to the clickup task

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -42,64 +43,57 @@ class ChannelAxis(Enum):
...
-            >>> FrameworkInfo(kernel_channels_mapping, {}, {})
+        >>> FrameworkInfo(kernel_channels_mapping, {}, {})
 
-            If an activation layer (tf.keras.layers.Activation) should be quantized and we know it's min/max outputs range in advanced, we can add it to activation_min_max_mapping for saving the statistics collection time. For example:
+        If an activation layer (tf.keras.layers.Activation) should be quantized and we know it's min/max outputs range in advanced, we can add it to activation_min_max_mapping for saving the statistics collection time. For example:
 
-            >>> activation_min_max_mapping = {'softmax': (0, 1)}
-            >>> FrameworkInfo(kernel_channels_mapping, activation_min_max_mapping, {})
+        >>> activation_min_max_mapping = {'softmax': (0, 1)}
+        >>> FrameworkInfo(kernel_channels_mapping, activation_min_max_mapping, {})
```
ofirgo: This example needs to be modified (I would consider removing it completely, fw info is no longer an external API)

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -42,64 +43,57 @@ class ChannelAxis(Enum):
...
-        self.activation_min_max_mapping = activation_min_max_mapping
-        self.layer_min_max_mapping = layer_min_max_mapping
-        self.kernel_ops_attributes_mapping = kernel_ops_attributes_mapping
-        self.out_channel_axis_mapping = out_channel_axis_mapping
+    activation_quantizer_mapping: Dict[QuantizationMethod, Callable]
+    kernel_channels_mapping: DefaultDict
+    activation_min_max_mapping: Dict[str, tuple]
+    layer_min_max_mapping: Dict[Any, tuple]
+    kernel_ops_attributes_mapping: DefaultDict
+    out_channel_axis_mapping: DefaultDict
```
ofirgo: Something to consider - I would get rid of the default dict and control the default return value in the functions.

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -42,64 +43,57 @@ class ChannelAxis(Enum):
...
-        """
+    """
 
-        self.activation_quantizer_mapping = activation_quantizer_mapping
-        self.kernel_channels_mapping = kernel_channels_mapping
-        self.activation_min_max_mapping = activation_min_max_mapping
-        self.layer_min_max_mapping = layer_min_max_mapping
-        self.kernel_ops_attributes_mapping = kernel_ops_attributes_mapping
-        self.out_channel_axis_mapping = out_channel_axis_mapping
+    activation_quantizer_mapping: Dict[QuantizationMethod, Callable]
```
ofirgo, line 69: make all these fields "protected" (add "_" prefix) to control accessing them only via the functions

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -88,6 +88,11 @@ def __init__(self,
         self.prior_info = None
         self.has_activation = has_activation
         self.is_custom = is_custom
+        fw_info = get_fw_info()
+        self.channel_axis = None if fw_info is None else fw_info.kernel_channels_mapping.get(layer_class)
```
ofirgo: not like this, access the value via a function (also for out_channel)

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -326,22 +328,19 @@ def get_num_parameters(self, fw_info) -> Tuple[int,int]:
...
-        Args:
-            fw_info: Framework info to decide which attributes should be quantized.
-
         Returns: Number of bytes the node's memory requires.
 
         """
         # TODO: this method is used for tensorboard only. If we want to enable logging of other attributes memory
         #  then it needs to be modified. But, it might be better to remove this method from the BaseNode completely.
-        kernel_attr = fw_info.get_kernel_op_attributes(self.type)[0]
+        kernel_attr = self.kernel_atts[0]
```
ofirgo: consider addressing the TODO here, I think it's the right time and I want us to clean the base_node a little bit during the refactor process

model_compression_toolkit/core/common/graph/functional_node.py
```diff
@@ -1,6 +1,21 @@
+# Copyright 2021 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo, line 1: if you added a missing license at least get the date right ;)

model_compression_toolkit/core/common/graph/functional_node.py
```diff
@@ -64,6 +80,12 @@ def __init__(self,
         self.functional_op = functional_op
         self.tensor_input_allocs = [] if tensor_input_allocs is None else tensor_input_allocs
 
+        fw_info = get_fw_info()
```
ofirgo: Why is it needed? it inherits from BaseNose, the attributes will be initialized there

model_compression_toolkit/core/common/mixed_precision/sensitivity_eval/metric_calculators.py
```diff
@@ -396,8 +393,8 @@ def get_output_nodes_for_metric(graph: Graph) -> List[BaseNode]:
         """
 
         return [n.node for n in graph.get_outputs()
-                if (graph.fw_info.is_kernel_op(n.node.type) and
-                    n.node.is_weights_quantization_enabled(graph.fw_info.get_kernel_op_attributes(n.node.type)[0])) or
+                if (get_fw_info().is_kernel_op(n.node.type) and
```
ofirgo: You shouldn't need to use fw info to check "is_kernel_op" now that you have the kernel information inside the node, this is just missing the point.
Am I missing anything?

model_compression_toolkit/core/keras/pruning/pruning_keras_implementation.py
```diff
@@ -161,21 +149,20 @@ def attrs_oi_channels_info_for_pruning(self,
...
 
         Returns:
             Dict[str, Tuple[int, int]]: A dictionary where each key is an attribute name (like 'kernel' or 'bias')
             and each value is a tuple representing the output and input channel axis indices respectively.
         """
 
         attributes_with_axis = {}
-        if fw_info.is_kernel_op(node.type):
-            kernel_attributes = fw_info.get_kernel_op_attributes(node.type)
+        if get_fw_info().is_kernel_op(node.type):
```
ofirgo: same, regarding usage of "is_kernel_op"

model_compression_toolkit/core/common/mixed_precision/sensitivity_eval/metric_calculators.py
elad-c: You're right. It's on the next commit after tests pass.

model_compression_toolkit/core/common/graph/functional_node.py
elad-c: Because I need `functional_op` and not `layer_class` in functional nodes. Another option is to move `self.functional_op = functional_op` above the `super()__init__`, and use `self.type` in `BaseNade`, but it seems like bad practice.

model_compression_toolkit/core/common/graph/functional_node.py
elad-c: The file was craeted in 2021...

model_compression_toolkit/ptq/pytorch/quantization_facade.py
elad-c: This import is only necessary if FOUND_TORCH is True, so no need to import it even though is in common.

model_compression_toolkit/ptq/pytorch/quantization_facade.py
elad-c: Separate commit

model_compression_toolkit/core/common/graph/functional_node.py
ofirgo: yes but the modifications that added the license happened today.
anyway I don't care I have no idea if the date matters to anyone

model_compression_toolkit/core/common/graph/functional_node.py
ofirgo: no you are right this is a bad idea since base node shouldn't be aware of functional_op.
another alternative can be to set these attribute vie a function _set_fw_node_attrs and call it once with layer_class (in base node) and once with functional_op.
not sure if it's much better, but I don't like the duplication..

think about it

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -88,6 +88,15 @@ def __init__(self,
         self.prior_info = None
         self.has_activation = has_activation
         self.is_custom = is_custom
+        self._set_fw_node_attrs(layer_class, framework_attr)
+
+    def _set_fw_node_attrs(self, node_type, framework_attr):
+        fw_info = get_fw_info()
+        self.channel_axis = None if fw_info is None else fw_info.get_kernel_channels(node_type)
```
irenaby: 1. Adding attributes dynamically outside init is bad practice and inconvenient for use.
2. You do not distinguish between attr not being set (due to fw_info being None), and attr being set to None (not applicable).
3. When can fw_info be None anyway and why do you allow it?

If you already handle it as a block, why not create a struct containing this info?

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -88,6 +88,15 @@ def __init__(self,
...
+        self.out_channel_axis = None if fw_info is None else fw_info.get_out_channel_axis(node_type)
+        self.minmax = None if fw_info is None else fw_info.get_layer_min_max(node_type, framework_attr)
+        self.kernel_atts = None if fw_info is None else fw_info.get_kernel_op_attributes(node_type)
```
irenaby: attrs?

irenaby: Also shouldn't we make it a single kernel_attr containing a string, instead of a list always assumed to be of length 1, if we agreed kernel means "kernel" and not "weight to quantize"

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -147,4 +137,57 @@ def activation_has_min_max(self, activation_name: str) -> bool:
...
+
+        Returns:
+            Node's output channel axis.
+
+        """
+        pass
+
+
+# Pointer to current FrameworkInfo class.
+_current_framework_info: Any = None
```
irenaby: its not Any. Also add type hints to functions below

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -42,64 +42,35 @@ class ChannelAxis(Enum):
...
+    """
 
-        self.activation_quantizer_mapping = activation_quantizer_mapping
-        self.kernel_channels_mapping = kernel_channels_mapping
-        self.activation_min_max_mapping = activation_min_max_mapping
-        self.layer_min_max_mapping = layer_min_max_mapping
-        self.kernel_ops_attributes_mapping = kernel_ops_attributes_mapping
-        self.out_channel_axis_mapping = out_channel_axis_mapping
+    activation_quantizer_mapping: Dict[QuantizationMethod, Callable]
+    kernel_channels_mapping: Dict
```
irenaby: add type hints to Dicts

model_compression_toolkit/core/common/substitutions/scale_equalization.py
```diff
@@ -77,7 +75,6 @@ def fixed_second_moment_after_relu(mu: np.ndarray,
 
 def scale_reshaping(scale: np.ndarray,
                     op2d: common.BaseNode,
-                    kernel_channel_mapping: DefaultDict,
                     kernel_str: str,
                     in_channels: bool = True) -> np.ndarray:
```
irenaby, line 79: should it be int?

model_compression_toolkit/core/keras/default_framework_info.py
```diff
@@ -16,99 +16,134 @@
...
+                                     DepthwiseConv2D: [DEPTHWISE_KERNEL],
+                                     Dense: [KERNEL],
+                                     Conv2DTranspose: [KERNEL]}
+
+    """
+    Map a layer to its kernel's output and input channels indices.
+    Map's values are tuples of (output_channel_index, input_channel_index).
+    Default value is returned for layers that are not included.
+    """
+    kernel_channels_mapping = {Conv2D: (3, 2),
```
irenaby: define NamedTuple with in and out channel

model_compression_toolkit/core/keras/default_framework_info.py
```diff
@@ -16,99 +16,134 @@
...
+    out_channel_axis_mapping = {Conv2D: -1,
+                                DepthwiseConv2D: -1,
+                                Dense: -1,
+                                Conv2DTranspose: -1}
+
+    """
+    Map from an activation function to its min/max output values (if known).
+    The values are used for tensor min/max values initialization.
+    """
+    activation_min_max_mapping = {SOFTMAX: (0, SOFTMAX_THRESHOLD),
```
irenaby: do we really need both activation minmax and layer minmax? And if yes, isnt this keras specific and has no business being in the base class?

model_compression_toolkit/core/pytorch/default_framework_info.py
```diff
@@ -12,87 +12,106 @@
...
+    """
+    out_channel_axis_mapping = {Conv2d: 1,
+                                Linear: -1,
+                                ConvTranspose2d: 1}
+
+    """
+    Map from an activation function to its min/max output values (if known).
+    The values are used for tensor min/max values initialization.
+    """
+    activation_min_max_mapping = {}  # should be an empty dict in Pytorch
```
irenaby: is it used by anyone?

model_compression_toolkit/core/pytorch/resource_utilization_data_facade.py
```diff
@@ -80,6 +81,7 @@ def pytorch_resource_utilization_data(in_model: Module,
             Logger.critical("Resource utilization data computation requires a MixedPrecisionQuantizationConfig object. "
                             "The provided 'mixed_precision_config' is not of this type.")
 
+        set_fw_info(PyTorchInfo)
```
irenaby: Instead of explicitly importing fw info object and setting it in all facades, you can just define a method set_torch_info that does it internally, like we talked.
I'll leave it to Ofir to argue about decorator

tests_pytest/pytorch_tests/torch_test_util/torch_test_mixin.py
```diff
@@ -34,7 +35,7 @@ class TorchFwMixin:
         TorchFooTester(BaseFooTester, TorchFwMixin):
           ...
     """
-    fw_info = DEFAULT_PYTORCH_INFO
+    set_fw_info(PyTorchInfo)
```
irenaby, line 38: a helper class should not automatically go and set global stuff when no one asked for it. and even if anyone asks for it it should be a patch, not changing the global state.

irenaby: Same for all tests that do it explicitly. Tests should not change the global state.

model_compression_toolkit/core/common/framework_info.py
irenaby: It should only contain private fields that are used by public implementation here. if the concrete class implements the public interface, no reason to define how it should do it. 
Also just to be clear, it is perfectly fine to have public fields without functional getters, not sure if Ofir meant everything or only those with functional api.

model_compression_toolkit/core/common/substitutions/scale_equalization.py
elad-c: what?

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -168,7 +165,7 @@ def get_out_channel_axis(cls, node_type: Any):
 
 
 # Pointer to current FrameworkInfo class.
-_current_framework_info: Any = None
+_current_framework_info: Union[type['PyTorchInfo'], type['KerasInfo']] = None
```
irenaby: Type[FrameworkInfo]

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -92,11 +103,74 @@ def __init__(self,
 
     def _set_fw_node_attrs(self, node_type, framework_attr):
         fw_info = get_fw_info()
-        self.channel_axis = None if fw_info is None else fw_info.get_kernel_channels(node_type)
-        self.out_channel_axis = None if fw_info is None else fw_info.get_out_channel_axis(node_type)
-        self.minmax = None if fw_info is None else fw_info.get_layer_min_max(node_type, framework_attr)
-        self.kernel_atts = None if fw_info is None else fw_info.get_kernel_op_attributes(node_type)
-        self.is_kernel_op = fw_info.is_kernel_op(node_type)
+        self.node_fw_info = None if fw_info is None else NodeFrameworkInfo(
```
irenaby: set the member in ctor: self.node_fw_info = self._get_fw_node_attrs(...)

model_compression_toolkit/core/common/framework_info.py
```diff
@@ -182,9 +179,12 @@ def get_fw_info():
     return _current_framework_info
 
 
-def set_fw_info(fw_info):
+def set_fw_info(fw_info: Union[type['PyTorchInfo'], type['KerasInfo']]):
```
irenaby: Type[FrameworkInfo]

tests_pytest/pytorch_tests/torch_test_util/torch_test_mixin.py
elad-c: Will fix this in next PR

irenaby (approved):

PR #1466, opened 2025-06-12
Apply activation quantization parameters selection

gouda-youichi:
Apply activation quantization parameters selection for Stage3-4.

### DIFF
--- a/model_compression_toolkit/core/common/graph/base_graph.py
+++ b/model_compression_toolkit/core/common/graph/base_graph.py
@@ -875,8 +875,7 @@ class Graph(nx.MultiDiGraph, GraphSearches):
                 def update(qc):
                     qc.activation_quantization_cfg = NodeActivationQuantizationConfig(fusing_op_quantization_cfg)
                     qc.activation_quantization_cfg.quant_mode = ActivationQuantizationMode.FLN_QUANT
-                node.quantization_cfg.update_all(update)
-                node.quantization_cfg.remove_duplicates()
+                node.quantization_cfg.update_all(update, remove_duplicates=True)
             else:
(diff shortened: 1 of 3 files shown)


model_compression_toolkit/core/common/quantization/quantization_params_generation/qparams_computation.py
```diff
@@ -135,7 +135,7 @@ def calculate_quantization_params(graph: Graph,
                     attr_cfg.weights_channels_axis = (output_channels_axis, attr_cfg.weights_channels_axis[1])
                     attr_cfg.set_weights_quantization_param(weights_params)
 
-            if n.is_activation_quantization_enabled():
+            if n.is_activation_quantization_enabled() or n.is_fln_quantization():
```
ofirgo: The condition here should be based on the fusing info, not the node candidate mode, similarly to the fix from PR #1462 
The same should apply for the condition in `set_activation_quantization_param` (the other change in this PR).

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -206,7 +206,7 @@ def set_activation_quantization_param(self,
             activation_params: Dictionary that contains weight quantization params.
 
         """
-        assert self.quant_mode == ActivationQuantizationMode.QUANT
+        assert self.quant_mode == ActivationQuantizationMode.QUANT or self.quant_mode == ActivationQuantizationMode.FLN_QUANT
```
ofirgo: this assertion needs to be modified to check for node is_enable_activation_quantization and same for fln quantization (similarly to all other places that this check occurs)

PR #1467, opened 2025-06-13
Apply fusinginfo quantization config to Activation quantization config.

gouda-youichi:
## Pull Request Description:
Apply fusinginfo quantization config to Activation quantization config. (For Stage3-3.5)


model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
```diff
@@ -89,13 +90,14 @@ def filter_node_candidates(node: BaseNode, fw_info) -> List[CandidateNodeQuantiz
...
 
-    if (kernel_attr is None or not node.is_weights_quantization_enabled(kernel_attr)) and not node.is_activation_quantization_enabled():
+    if (kernel_attr is None or not node.is_weights_quantization_enabled(kernel_attr)) and (not node.is_activation_quantization_enabled() and not node.is_fln_quantization()):
         # If activation quantization is disabled and the node doesn't have a kernel or doesn't quantize the kernel,
         # but for some reason the node has multiple candidates then replace it with a single dummy candidate with
         # default bit-width values.
         single_dummy_candidate = filtered_candidates[0]
         single_dummy_candidate.activation_quantization_cfg.activation_n_bits = FLOAT_BITWIDTH
         single_dummy_candidate.activation_quantization_cfg.activation_quantization_method = QuantizationMethod.POWER_OF_TWO
+        single_dummy_candidate.activation_quantization_cfg.activation_quantization_params_fn = get_activation_quantization_params_fn(QuantizationMethod.POWER_OF_TWO)
```
ofirgo: why would you need `quantization_params_fn` if this is a case of a node that is **not** quantizing activations?

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
```diff
@@ -89,13 +90,14 @@ def filter_node_candidates(node: BaseNode, fw_info) -> List[CandidateNodeQuantiz
     final_candidates = copy.deepcopy(node.candidates_quantization_cfg)
     kernel_attr = fw_info.get_kernel_op_attributes(node.type)[0]
 
-    if (kernel_attr is None or not node.is_weights_quantization_enabled(kernel_attr)) and not node.is_activation_quantization_enabled():
+    if (kernel_attr is None or not node.is_weights_quantization_enabled(kernel_attr)) and (not node.is_activation_quantization_enabled() and not node.is_fln_quantization()):
```
ofirgo: I think that you can just check the activation condition here to check whether the mode is ActivationQuantizationMode.NO_QUANT, because this is the only case where you want to modify the candidates. Correct me if I'm overlooking anything here

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
```diff
@@ -104,9 +106,10 @@ def filter_node_candidates(node: BaseNode, fw_info) -> List[CandidateNodeQuantiz
 
         final_candidates = [single_dummy_candidate]
 
-    elif not node.is_activation_quantization_enabled():
+    elif not node.is_activation_quantization_enabled() and not node.is_fln_quantization():
```
ofirgo: again, I hink the condition can be simplyfied here to check for ActivationQuantizationMode.NO_QUANT

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
```diff
@@ -115,6 +118,7 @@ def filter_node_candidates(node: BaseNode, fw_info) -> List[CandidateNodeQuantiz
         for c in filtered_candidates:
             c.activation_quantization_cfg.activation_n_bits = FLOAT_BITWIDTH
             c.activation_quantization_cfg.activation_quantization_method = QuantizationMethod.POWER_OF_TWO
+            c.activation_quantization_cfg.activation_quantization_params_fn = get_activation_quantization_params_fn(QuantizationMethod.POWER_OF_TWO)
```
ofirgo: again, I don't understand why should we modify config of an FLN. If it is fln no_quant it should be set when setting fusing info on the graph

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -907,11 +908,23 @@ def disable_fused_nodes_activation_quantization(self):
         """
```
ofirgo, line 880: please change this function name to "override_fused_node_activation_quantization_candidates"

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -907,11 +908,23 @@ def disable_fused_nodes_activation_quantization(self):
...
         Disable activation quantization for all nodes in fused operations,
         except for the last node in each fused group.
+        Update the value of quantization_config with the value of op_quaitization_cfg from FusingInfo.
         """
         nodes_to_disable = self.fusing_info.get_inner_fln_nodes()
         for node in nodes_to_disable:
+            fused_node_op_id = self.fusing_info.get_fused_op_id_for_node(node.name)
+            fusiong_op_quaitization_cfg = self.fusing_info.get_fused_op_quantization_config(fused_node_op_id) 
             for qc in node.candidates_quantization_cfg:
```
ofirgo: the modification should override the entire existing candidate list and replace it with a single, newly initialized candidate from the op_cfg.

tests_pytest/_test_util/graph_builder_utils.py
```diff
@@ -105,7 +106,7 @@ def build_nbits_qc(a_nbits=8, a_enable=True, w_attr=None, pos_attr=(32, False, (
         enable_activation_quantization=a_enable,
         default_weight_attr_config=AttributeQuantizationConfig(weights_n_bits=pos_attr[0],
                                                                enable_weights_quantization=pos_attr[1]),
-        activation_quantization_method=QuantizationMethod.POWER_OF_TWO,
+        activation_quantization_method=activation_quantization_method,
```
ofirgo: this is a test util function that is used all over the tests package. changing it might change the behavior of a test to something that the developer that wrote the test might not intend.
Therefore, it is preferred not to modify those functions, even by duplicating them and extending their behavior for your specific needs.

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
+# Copyright 2025 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo: this test file should be under /unit_tests/core/graph

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+        for test_c, exp_c in zip(candidates, exp_candidates):
+            test_actq_cfg = test_c.activation_quantization_cfg
+            exp_actq_cfg = exp_c.activation_quantization_cfg
+            assert test_actq_cfg.quant_mode == exp_actq_cfg.quant_mode
+            assert test_actq_cfg.activation_n_bits == exp_actq_cfg.activation_n_bits
+            assert test_actq_cfg.activation_quantization_method == exp_actq_cfg.activation_quantization_method
+            assert test_actq_cfg.activation_quantization_params_fn == exp_actq_cfg.activation_quantization_params_fn
+            assert test_actq_cfg.signedness == exp_actq_cfg.signedness
+
+    def test_disable_fused_nodes_activation_quantization(self, fusing_info_generator_with_qconfig):
```
ofirgo: I'm having a hard time understanding all the setup for this test.
It is a unit test, it is supposed to be small and check dedicated behaviour.
try to simplify it by using mocks for the specific functionality that is used when calling the tested function - you don't need a real graph, a real fusing info or a real tpc.

tests_pytest/common_tests/unit_tests/test_filter_nodes_candidates.py
```diff
@@ -0,0 +1,247 @@
+# Copyright 2025 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo: find the correct location of this test file according to the source code location of the tested function

tests_pytest/common_tests/unit_tests/test_filter_nodes_candidates.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        for test_c, exp_c in zip(candidates, exp_candidates):
+            test_actq_cfg = test_c.activation_quantization_cfg
+            exp_actq_cfg = exp_c.activation_quantization_cfg
+            assert test_actq_cfg.activation_n_bits == exp_actq_cfg.activation_n_bits
+            assert test_actq_cfg.activation_quantization_method == exp_actq_cfg.activation_quantization_method
+            assert test_actq_cfg.activation_quantization_params_fn == exp_actq_cfg.activation_quantization_params_fn
+            assert test_actq_cfg.signedness == exp_actq_cfg.signedness
+
+    def test_filter_nodes_candidates_updates_node_activation_quantization_config(self, fusing_info_generator_with_qconfig):
```
ofirgo: here as well the preparation for the test is exhasting, it makes the test unreadable and very complicated to understand what is tested and what is the expected behaviour. please try to simplify it.

ofirgo (changes requested):

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
gouda-youichi: Since there are nodes in the graph that are in PRESERVE_QUANT mode, we believe that judging only by NO_QUANT is insufficient.

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
gouda-youichi: You're right.
This is a remnant of an implementation that originally assumed that FLN_QUANT mode would not be used. I will delete it.

tests_pytest/common_tests/unit_tests/test_base_graph.py
gouda-youichi: I would like you to explain in more detail.
The target of this test is the member function disable_fused_nodes_activation_quantization() of the Graph class, so I think it is appropriate to use the real Graph class. I also thought it would be appropriate to refer to the real implementation of fusing_info. Could you tell me specifically what kind of test sequence you have in mind?

tests_pytest/common_tests/unit_tests/test_filter_nodes_candidates.py
gouda-youichi: I understand. I will simplify by changing the test target to filter_node_candidates(), which processes on a node-by-node basis.

tests_pytest/_test_util/graph_builder_utils.py
gouda-youichi: I understand. I will not change the original function, but will change it so that any additional processing that is required is implemented in a closed form within this test.

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
gouda-youichi: Similar as below.
https://github.com/sony/model_optimization/pull/1467#discussion_r2148934415
image

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
gouda-youichi: Same as below.
https://github.com/sony/model_optimization/pull/1467#discussion_r2148942396
image

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
ofirgo: in case of preserving you also don't need to touch the candidates, so it's not a problem. That is, because the candidates of a preserving node are actually the candidates of the node from which he preserved

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
ofirgo: same answer :)

tests_pytest/common_tests/unit_tests/test_base_graph.py
ofirgo: The test you're describing is more of an integration test, which takes several dependent modules and checks how they're integrated together.
The purpose of a unit test is to test a simple functionality independently, assuming that all other dependencies behave in a certain way - we want to verify that our function returns the expected output.
In this case, for example, you need to mock FusingInfo's `get_inner_fln_nodes` method, which would return a node to test, `get_fused_op_id_for_node`, which would return an op_id for testing, etc.
You don't need a real graph or real fusing info for that.

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+from tests.common_tests.helpers.generate_test_tpc import generate_test_attr_configs, generate_test_op_qc
+from tests_pytest._test_util.graph_builder_utils import build_nbits_qc as build_qc
+from tests_pytest.common_tests.unit_tests.test_filter_nodes_candidates import create_mock_base_node
+from model_compression_toolkit.core.common.quantization.node_quantization_config import ActivationQuantizationMode
+import copy
+
+# Setup TEST_QC and TEST_QCO for testing.
+TEST_QC = generate_test_op_qc(**generate_test_attr_configs(),
+                              activation_n_bits=4,
+                              activation_quantization_method=QuantizationMethod.LUT_POT_QUANTIZER)
```
ofirgo: Consider replacing the activation_quantization_method to POWER_OF_TWO, which is more common. it is not relevant for what you're testing, but the chocie of LUT_POT_QUANTIZER is unclear and makes you wonder whether it is a part of the test or not

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+from model_compression_toolkit.core.common.quantization.quantization_params_generation.symmetric_selection import symmetric_selection_histogram
+from model_compression_toolkit.target_platform_capabilities.schema.mct_current_schema import Signedness
+from tests.common_tests.helpers.generate_test_tpc import generate_test_attr_configs, generate_test_op_qc
+from tests_pytest._test_util.graph_builder_utils import build_nbits_qc as build_qc
+from tests_pytest.common_tests.unit_tests.test_filter_nodes_candidates import create_mock_base_node
+from model_compression_toolkit.core.common.quantization.node_quantization_config import ActivationQuantizationMode
+import copy
+
+# Setup TEST_QC and TEST_QCO for testing.
+TEST_QC = generate_test_op_qc(**generate_test_attr_configs(),
```
ofirgo: you can initialize this object under `fusing_patterns_with_qconfig` it doesn't need to be global since it is not used anywhere else

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+                              activation_n_bits=4,
```
ofirgo: please change this to 16, it is not relevant for the test but it mimics the actual common scenario where FLN quantization would be to 16 bit

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+                              activation_quantization_method=QuantizationMethod.LUT_POT_QUANTIZER)
+
+
+@pytest.fixture
+def fusing_patterns_with_qconfig():
```
ofirgo: not sure if it is necessary to define this as a fixture. also, this can be part of `fusing_info_generator_with_qconfig` since it is only used there.

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+    def setup_mock_graph_and_nodes(self):
+        """
+        Set up mock objects for testing the disable_fused_nodes_activation_quantization modules.
+
+        Mocks the behavior of nodes in a graph that support activation and weight quantization settings.
+        Creates a graph containing nodes with multiple quantization candidates
+        to test whether the expected configurations are set correctly.
+        """
+
+        self.fw_info = Mock()
```
ofirgo: you can set this mock directly in the test no need for the duplication of setting it here and then reading it into a local `fw_info=self.fw_info` later.

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+            assert test_actq_cfg.activation_quantization_params_fn == exp_actq_cfg.activation_quantization_params_fn
+            assert test_actq_cfg.signedness == exp_actq_cfg.signedness
+
+    def test_disable_fused_nodes_activation_quantization(self, fusing_info_generator_with_qconfig):
+        """
+        Test the disable_fused_nodes_activation_quantizatio function for a graph with multiple nodes and configurations.
+        """
+        graph = self.graph
+        fw_info = self.fw_info
+        fw_info.get_kernel_op_attributes.return_value = ['weight']
```
ofirgo: is this mock functionality needed for the test? if `get_kernel_op_attributes` is not called in the inner function that runs during the test then no need to set this return value

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+        
+        graph.fw_info = fw_info
```
ofirgo: not sure it is needed (unless `graph.fw_info` is being called in the actual tested function)

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+        graph.fusing_info = fusing_info_generator_with_qconfig.generate_fusing_info(graph)
```
ofirgo: I don't think that you need to initialize the entire fusing_info, just define it as a mock and set the behavior of the function you need

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+        flatten_qc_cfg = [create_exp_candidate_cfg(exp_candidate_base1, **exp_actq_cfg_params_dict3)]
+        ### Expected values when unchanged
+        linear_qc_cfg = [exp_candidate_base1, exp_candidate_base2]
+
+        self.exp_node_config_list = [conv_1_qc_cfg, relu_1_qc_cfg,
+                                     conv_2_qc_cfg, tanh_qc_cfg,
+                                     conv_3_qc_cfg, relu_2_qc_cfg,
+                                     flatten_qc_cfg, linear_qc_cfg]
+    
+    def check_candidates_activation_qcfg(self, candidates, exp_candidates):
```
ofirgo: This check needs a little modification - first of all, if you are expecting an FLN quantization, check that only a single candidate is set for the node, if not an FLN - verify the number of expected candidates.
Then, we only care to verify the content of the candidate for an FLN node, so no need for the for loop in my opinion

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+                            nodes=mock_nodes_list,
+                            output_nodes=[OutTensor(mock_nodes_list[5], 0)],
+                            edge_list=edges_list)
+
+    @pytest.fixture(autouse=True)
+    def expected_candidates_fixture(self):
+        """
+        Set up and create expected values and candidates for tests related to the disable_fused_nodes_activation_quantization modules.
+        """
+        def create_exp_candidate_cfg(candidate, quant_mode, n_bits, actq_params_fn, actq_method, signedness=None):
```
ofirgo: I suggest you split this setup function, extract `create_exp_candidate_cfg` to an outer scope (no need for nested function here) and maybe create a function that gets the arguments to build and set a config, instead of working with these long dictionaries.
All this is just for simplification and readability.

tests_pytest/common_tests/unit_tests/test_base_graph.py
```diff
@@ -0,0 +1,215 @@
...
+
+        ### expected is test_disable_fused_nodes_activation_quantization 
+        exp_candidate_base1 = build_qc(a_nbits=8, a_enable=True, w_attr={'weight': (8, True)},
+                                        activation_quantization_fn=symmetric_selection_histogram,
+                                        activation_quantization_method=QuantizationMethod.SYMMETRIC)
+        exp_candidate_base2 = build_qc(a_nbits=4, a_enable=True, w_attr={'weight': (4, True)},
+                                        activation_quantization_fn=symmetric_selection_histogram,
+                                        activation_quantization_method=QuantizationMethod.SYMMETRIC)
+        
+        exp_actq_cfg_params_dict1 = {'quant_mode': ActivationQuantizationMode.FLN_QUANT,
```
ofirgo: no need to include lut here, this just makes it more difficult to understand

tests_pytest/common_tests/unit_tests/test_base_graph.py
ofirgo: consider using `@pytest.mark.parametrize` for different test cases instead of looping over all the cases

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
gouda-youichi: I'm modifing by using is_no_quant and is_fln_no_quant.

ofirgo: why do you need to check is_fln_no_quant? the only case in which you want to enter this "if" is if the node is defined as "no quantization" any other case - changing the candidates is unnecessary.
If you encountered some issue when putting just "no quant" in this condition please let me know

tests_pytest/common_tests/unit_tests/core/graph/test_base_graph.py
```diff
@@ -0,0 +1,96 @@
...
+                              activation_n_bits=16,
+                              activation_quantization_method=QuantizationMethod.POWER_OF_TWO)
+
+def build_mock_fusing_info(nodes):
+    """
+    Creates a mock FusingInfo object that simulates the behavior of fusing information in a graph.
+    """
+
+    fusing_info = Mock(spec=FusingInfo)
+    fusing_info.get_inner_fln_nodes.return_value = [nodes[0], nodes[1]]
```
ofirgo, line 45: I suggest to split this into two test ssenarios - 
1. FLN quant wuth expected quantization
2. FLN with expected quantization disabled

also, please add a case where the OpQuantizationConfig is not None, but activation quantization is disabled in the config (should get "FLN_NO_QUANT" in the test)

tests_pytest/common_tests/unit_tests/core/graph/test_base_graph.py
```diff
@@ -0,0 +1,96 @@
...
+from model_compression_toolkit.core.common.fusion.fusing_info import FusingInfo
+from model_compression_toolkit.target_platform_capabilities.schema.mct_current_schema import Signedness
+from tests.common_tests.helpers.generate_test_tpc import generate_test_attr_configs, generate_test_op_qc
+from model_compression_toolkit.core.common.quantization.node_quantization_config import ActivationQuantizationMode, NodeActivationQuantizationConfig
+from model_compression_toolkit.core.common.quantization.candidate_node_quantization_config import CandidateNodeQuantizationConfig
+from model_compression_toolkit.core.common.quantization.quantization_params_generation.power_of_two_selection import power_of_two_selection_histogram
+
+    
+# Setup TEST_QC and TEST_QCO for testing.
+TEST_QC = generate_test_op_qc(**generate_test_attr_configs(),
```
ofirgo: I'm not sure why you need this `generate_test_op_qc` and `generate_test_attr_configs` for this test. just create an OpQuantizationConfig with only the relevant attributes (which I believe are activation_quantization_method, activation_n_bits and enable_activation_quantization).

tests_pytest/_test_util/graph_builder_utils.py
```diff
@@ -139,3 +139,61 @@ def build_nbits_qc(a_nbits=8, a_enable=True, w_attr=None, pos_attr=(32, False, (
         assert qc.weights_quantization_cfg.get_attr_config(pos).enable_weights_quantization == pos_attr[1]
 
     return qc
+
+def build_nbits_qc_with_quantization_method(a_nbits=8, a_enable=True, w_attr=None, pos_attr=(32, False, ()),
```
ofirgo: are you using this function somewhere in your tests? if not, please remove it. otherwise, maybe think of a way to prevent this duplication, if it is only necessary for a single test there must be a better way to create the qc

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -903,15 +904,29 @@ def _find_intermediate_and_exit_nodes(self, entry_node: BaseNode, fw_impl: Any)
 
         return intermediate_nodes, next_node
 
-    def disable_fused_nodes_activation_quantization(self):
+    def override_fused_node_activation_quantization_candidates(self):
```
ofirgo, line 879: how is there no change for this method usages? anywhere that uses `disable_fused_nodes_activation_quantization` should be modified to use `override_fused_node_activation_quantization_candidates`

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -903,15 +904,29 @@ def _find_intermediate_and_exit_nodes(self, entry_node: BaseNode, fw_impl: Any)
...
         """
         Disable activation quantization for all nodes in fused operations,
         except for the last node in each fused group.
+        Update the value of quantization_config with the value of op_quaitization_cfg from FusingInfo.
```
ofirgo, line 883: this sentence should replace the first one - "Disable activation..."

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -903,15 +904,29 @@ def _find_intermediate_and_exit_nodes(self, entry_node: BaseNode, fw_impl: Any)
...
         """
         nodes_to_disable = self.fusing_info.get_inner_fln_nodes()
         for node in nodes_to_disable:
-            for qc in node.candidates_quantization_cfg:
-                qc.activation_quantization_cfg.quant_mode = ActivationQuantizationMode.FLN_QUANT
+            fused_node_op_id = self.fusing_info.get_fused_op_id_for_node(node.name)
+            fusiong_op_quaitization_cfg = self.fusing_info.get_fused_op_quantization_config(fused_node_op_id) 
+            if fusiong_op_quaitization_cfg is not None:
+                for qc in node.candidates_quantization_cfg:
```
ofirgo, line 899: as we previously discussed, there shouldn't be an inplace modification to each candidate here, but instead, this method should override the candidates of an fln node with a list that include 1 new candidates, with its attribute taken from `fusiong_op_quaitization_cfg`

ofirgo (changes requested):

tests_pytest/common_tests/unit_tests/test_filter_nodes_candidates.py
gouda-youichi: move to tests_pytest/common_tests/unit_tests/core/quantization/test_filter_nodes_candidates.py

tests_pytest/common_tests/unit_tests/core/graph/test_base_graph.py
gouda-youichi: I parameterized the test and added an if statement condition to three patterns.

tests_pytest/common_tests/unit_tests/core/graph/test_base_graph.py
gouda-youichi: I changed OpQuantizationConfig to Mock.

tests_pytest/_test_util/graph_builder_utils.py
gouda-youichi: I delete that function.

model_compression_toolkit/core/common/graph/base_graph.py
gouda-youichi: I change the function name.

model_compression_toolkit/core/common/graph/base_graph.py
gouda-youichi: I correct that comment.

model_compression_toolkit/core/common/graph/base_graph.py
gouda-youichi: I removed the for statement and changed it so that only the first element is changed.

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -876,15 +877,28 @@ def _find_intermediate_and_exit_nodes(self, entry_node: BaseNode, fw_impl: Any)
...
 
-    def disable_fused_nodes_activation_quantization(self):
+    def override_fused_node_activation_quantization_candidates(self):
         """
-        Disable activation quantization for all nodes in fused operations,
+        Override fused node activation quantization candidates for all nodes in fused operations,
         except for the last node in each fused group.
+        Update the value of quantization_config with the value of op_quaitization_cfg from FusingInfo.
         """
         nodes_to_disable = self.fusing_info.get_inner_fln_nodes()
```
ofirgo: change the variable name, it is no longer nodes_to_disable, it is nodes_in_fln

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -876,15 +877,28 @@ def _find_intermediate_and_exit_nodes(self, entry_node: BaseNode, fw_impl: Any)
...
         for node in nodes_to_disable:
-            for qc in node.candidates_quantization_cfg:
-                qc.activation_quantization_cfg.quant_mode = ActivationQuantizationMode.FLN_QUANT
+            fused_node_op_id = self.fusing_info.get_fused_op_id_for_node(node.name)
+            fusiong_op_quaitization_cfg = self.fusing_info.get_fused_op_quantization_config(fused_node_op_id) 
+            if fusiong_op_quaitization_cfg is not None and fusiong_op_quaitization_cfg.enable_activation_quantization:
+                # Set ActivationQuantizationMode to FLN_QUANT and update the value of quantization_config
+                actq_cfg = node.candidates_quantization_cfg[0].activation_quantization_cfg
```
ofirgo: This is still not completed.
you shouldn't override the existing candidate at index 0 (or all candidates in the "else" part).
What you should do is remove all candidates, and initialize a new one.
The new activation config is taken from the fusing info.
The new weights config can be similar to the deleted candidate at index 0 (it doesn't matter for now)

ofirgo: You can initialize a `CandidateNodeQuantizationConfig` object by passing it `activation_quantization_cfg: NodeActivationQuantizationConfig` and `weights_quantization_cfg: NodeWeightsQuantizationConfig`.
Something like:
```
org_candidate = node.candidates_quantization_cfg[0]
activation_quantization_cfg = NodeActivationQuantizationConfig(op_cfg=fusiong_op_quaitization_cfg, org_candidate.activation_quantization_cfg.activation_quantization_fn, activation_quantization_params_fn=org_candidate.activation_quantization_cfg.activation_quantization_params_fn )
weights_quantization_cfg = org_candidate.weights_quantization_cfg

node.candidates_quantization_cfg = []
node.candidates_quantization_cfg.append(CandidateNodeQuantizationConfig(activation_quantization_cfg=activation_quantization_cfg, weights_quantization_cfg=weights_quantization_cfg))
```

and then set quant_mode to FLN

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
gouda-youichi: I understand. I corrected by using is_no_quant.

model_compression_toolkit/core/common/graph/base_graph.py
gouda-youichi: The proposed code uses CandidateNodeQuantizationConfig, so I need to import it, but if I import it at the top of the file, I get a circular reference error.
The error can be avoided by lazily importing it within the function. like below:
Is it right?

```
    def override_fused_node_activation_quantization_candidates(self):
        from model_compression_toolkit.core.common.quantization.candidate_node_quantization_config import CandidateNodeQuantizationConfig
        nodes_in_fln = self.fusing_info.get_inner_fln_nodes()
        for node in nodes_in_fln:
[Omitted]
                node.candidates_quantization_cfg.append(CandidateNodeQuantizationConfig(activation_quantization_cfg=activation_quantization_cfg, weights_quantization_cfg=weights_quantization_cfg))
```

ofirgo: Try to import it at the top under "TYPE_CHECKING" scope like this:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING: 
        from model_compression_toolkit.core.common.quantization.candidate_node_quantization_config import CandidateNodeQuantizationConfig
```

if it doesn't solve the circular error, then you can import it inside the function

model_compression_toolkit/core/common/graph/base_graph.py
gouda-youichi: I fixed it.

model_compression_toolkit/core/common/graph/base_graph.py
gouda-youichi: I changed the variable name.

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -876,15 +876,32 @@ def _find_intermediate_and_exit_nodes(self, entry_node: BaseNode, fw_impl: Any)
...
+                activation_quantization_cfg = NodeActivationQuantizationConfig(qc=org_candidate,
+                                                                               op_cfg=fusiong_op_quaitization_cfg, 
+                                                                               activation_quantization_fn=org_candidate.activation_quantization_cfg.activation_quantization_fn,
+                                                                               activation_quantization_params_fn=org_candidate.activation_quantization_cfg.activation_quantization_params_fn)
+                weights_quantization_cfg = org_candidate.weights_quantization_cfg
+                node.candidates_quantization_cfg = []
+                node.candidates_quantization_cfg.append(CandidateNodeQuantizationConfig(activation_quantization_cfg=activation_quantization_cfg, weights_quantization_cfg=weights_quantization_cfg))
+                activation_quantization_cfg.quant_mode = ActivationQuantizationMode.FLN_QUANT
+            else:
+                for qc in node.candidates_quantization_cfg:
```
ofirgo, line 903: last thing regarding this function - you should remove the for loop from the "else" part as well.
change it so the single new candidate is initialized and append to an empty list, and then just mofidy the quant mode to either "FLN_QUANT" or "FLN_NO_QUANT" based on the if

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -876,15 +876,32 @@ def _find_intermediate_and_exit_nodes(self, entry_node: BaseNode, fw_impl: Any)
...
+        """
+        from model_compression_toolkit.core.common.quantization.candidate_node_quantization_config import CandidateNodeQuantizationConfig
+
+        nodes_in_fln = self.fusing_info.get_inner_fln_nodes()
+        for node in nodes_in_fln:
+            fused_node_op_id = self.fusing_info.get_fused_op_id_for_node(node.name)
+            fusiong_op_quaitization_cfg = self.fusing_info.get_fused_op_quantization_config(fused_node_op_id) 
+            if fusiong_op_quaitization_cfg is not None and fusiong_op_quaitization_cfg.enable_activation_quantization:
+                # Set ActivationQuantizationMode to FLN_QUANT and update the value of quantization_config
+                org_candidate = node.candidates_quantization_cfg[0]
```
ofirgo: Sorry for missleading you regarding this function.
The situation is as follows:
Each candidates contains both weights and activation configuration. while for FLN nodes (quantized or not quantized) we should only have a single **activation** config, as specified in the tpc, **the weights candidates should remain the same**.
That is why you are supposed to loop over the candidates like you did before, just override only the activation config.
something like that:
```
org_candidate = node.candidates_quantization_cfg[0]
if fusiong_op_quaitization_cfg is not None and fusiong_op_quaitization_cfg.enable_activation_quantization:
    # Set ActivationQuantizationMode to FLN_QUANT and update the value of quantization_config
    activation_quantization_cfg = NodeActivationQuantizationConfig(qc=org_candidate,
                                                                   op_cfg=fusiong_op_quaitization_cfg,
                                                                   activation_quantization_fn=org_candidate.activation_quantization_cfg.activation_quantization_fn,
                                                                   activation_quantization_params_fn=org_candidate.activation_quantization_cfg.activation_quantization_params_fn)
    activation_quantization_cfg.quant_mode = ActivationQuantizationMode.FLN_QUANT
    for c in node.candidates_quantization_cfg:
        c.activation_quantization_cfg = activation_quantization_cfg
else:
    for c in node.candidates_quantization_cfg:
        c.activation_quantization_cfg.quant_mode = ActivationQuantizationMode.FLN_NO_QUANT
```

ofirgo: This will also require a modification to the filter_node_candidates. See my other comment

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
```diff
@@ -102,9 +101,10 @@ def filter_node_candidates(node: BaseNode) -> List[CandidateNodeQuantizationConf
 
         final_candidates = [single_dummy_candidate]
 
-    elif not node.is_activation_quantization_enabled():
+    elif node.is_no_quantization():
```
ofirgo, line 104: Following the understanding regarding the override method, now we can have the following situation:
An FLN node with multiple weights candidates, but a single activation (fln) candidate. In this case we need to filter duplicate candidates, but the durrent function doesn't take care of it.
I think that the correct solution would be to add another `elif` statemeant as follows:
```
    elif node.is_fln_no_quantization() or node.is_fln_quantization():
        seen_candidates = set()
        filtered_candidates = [candidate for candidate in filtered_candidates if
                               candidate.weights_quantization_cfg not in seen_candidates
                               and not seen_candidates.add(candidate.weights_quantization_cfg)]
        final_candidates = _filter_bit_method_dups(filtered_candidates, node.kernel_attr)
```

ofirgo (approved):

PR #1475, opened 2025-06-23
Initial quantization preparation

irenaby:
## Pull Request Description:
Initial quantization preparation module.
New module quantization_preparation was added, with a single entry point load_fqc_configuration, which sets nodes quantization candidates per FQC and graph's Fusion info. Most of the relevant code was moved / rewritten there. 
Major change: node quantization candidates (activation, weights and attribute configs only contain info from FQC and do not read QuantizationConfig anymore. This breaks lots of code elsewhere, so as a first stage all missing info is injected back right after quantization_preparation, and candidates are filtered for mp/no mp.
Set manual bitwidth was rewritten with better TPC compatability validation, and is done on nodes quantization configs instead of on TPC objects as previously.

Main gaps to be addressed in next PRs:
Move/rewrite filter_node_qco_by_graph from set_node_qantization_config to the new module, add testing for the module.
Manual bitwidth testing was partly fixed, mostly disabled.
Remove the injection of info from QuantizationConfig and filtering for mp / no mp.


model_compression_toolkit/core/common/framework_info.py
```diff
@@ -85,20 +79,7 @@ def get_kernel_op_attribute(cls, node_type: Any) -> str:
         Returns:
             Attribute the layer has and should be quantized.
         """
-        return cls.kernel_ops_attribute_mapping.get(node_type, DEFAULT_KERNEL_ATTRIBUTE)
-
-    @classmethod
-    def is_kernel_op(cls, node_type: Any) -> bool:
```
ofirgo, line 91: What's replacing this?

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,212 @@
...
+
+    for n in graph.get_topo_sorted_nodes():
+        set_quantization_configs_to_node(node=n,
+                                         graph=graph,
+                                         fqc=fqc)
+    return graph
+
+
+def set_fusion_info(graph, fqc) -> Graph:
+    # TODO fix the horrible dict with const keys inside get_fusing_patterns. use named tuple or class
```
ofirgo: please add this as subtask to the clickup item of this refactor (and link to here)

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,212 @@
...
+    Args:
+        weight_channel_axis: channels axes of the node's kernel.
+        op_cfg: quantization config for the op.
+        node_attrs_list: A list of the node's weights attributes names.
+
+    Returns:
+        Candidate quantization config.
+    """
+
+    # TODO irena: i think we shouldn't inject methods here, it's quantization implementation, not configuration
```
ofirgo, line 211: Can you please explain what you mean? I'm not sure I understand

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,212 @@
...
+
+    node_attrs_list = node.get_node_weights_attributes()
+    mp_candidates = [_create_candidate(node.channel_axis, qc, node_attrs_list)
+                     for qc in candidates_qcs]
+    sp_cfg = _create_candidate(node.channel_axis, base_config, node_attrs_list)
+
+    node.tpc_quantization_info = TPCQuantizationInfo(base_quantization_cfg=sp_cfg,
+                                                     candidates_quantization_cfg=mp_candidates)
+
+    # TODO is not needed anymore as find min/max candidate look for a real max/min, but some tests still count on it
```
ofirgo, line 73: Let's handle the tests in a separate PR, please add this to the clickup item as well (and remove this method once it's done?)

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,212 @@
...
+    node.sort_node_candidates()
+
+    if not node.has_activation:
+        node.tpc_quantization_info.update_activation_quantization_mode(ActivationQuantizationMode.NO_QUANT)
+
+    _disable_unsupported_quant_preserving(node, graph)
+
+
+def _disable_unsupported_quant_preserving(node, graph):
```
ofirgo: why do we want to do this?
also, please add method description maybe I just don't get what it supposed to do.

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,212 @@
...
+        return filter_qcos[0]
+
+    if qcos:
+        return qcos[0]
+
+    # TODO irena move check for custom from graph.set_fqc here
+    return fqc.tpc.default_qco if return_default else None
+
+
+# TODO irena copied from graph.set_fqc here.
```
ofirgo: can it be removed from there?

model_compression_toolkit/core/common/quantization/candidate_node_quantization_config.py
```diff
@@ -12,73 +12,120 @@
...
+@dataclass(eq=True)
 class CandidateNodeQuantizationConfig(BaseNodeQuantizationConfig):
     """
-    Class for representing candidate node configuration, which includes weights and activation configuration combined.
+    Candidate quantization configuration for a node.
     """
+    activation_quantization_cfg: NodeActivationQuantizationConfig
+    # TODO irena: None is passed in several places, need to check if it's handled properly or it's only passed in cases
+    #  that do not affect anything (my guess is it's the second).
+    #  I think in general it makes more sense to set it to None when there are no weights, and maybe when all weights
```
ofirgo, line 31: This is a major design choice. It might be a good idea but I don't think we should do this as part of this phase of the refactor.

model_compression_toolkit/core/common/quantization/candidate_node_quantization_config.py
```diff
@@ -12,73 +12,120 @@
...
+    #  are unquantized, and handle it properly everywhere.
+    weights_quantization_cfg: Optional[NodeWeightsQuantizationConfig]
+
+
+# TODO irena: currently all code still looks at candidates_quantization_cfg as previously, so this is just an initial
+#  implementation. For now base config is completely separated from candidates (base config must be equal to one of the
+#  candidates, but we create a separate copy), and updating in place is allowed. Also we require quantization mode to
+#  be identical between all configs.
+@dataclass
+class TPCQuantizationInfo:
```
ofirgo: I don't like this name, especially when it is projected in the objects above it (node).
TPC is meant to describe platform capabilities, this is a node quantization setup, the fact that it comes from the TPC is irrelevant imo.
I would suggest to keep this in the region of "quantization config", "quantization candidate" etc.

model_compression_toolkit/core/common/quantization/filter_nodes_candidates.py
```diff
@@ -34,7 +34,7 @@ def filter_nodes_candidates(graph: Graph):
     """
     nodes = list(graph.nodes)
     for n in nodes:
-        n.candidates_quantization_cfg = filter_node_candidates(node=n)
+        n.tpc_quantization_info.candidates_quantization_cfg = filter_node_candidates(node=n)
```
ofirgo: consider implementing a dedicated "API" for modifying node's candidates list, instead of accessing through n.tpc_quantization_info

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -95,7 +90,7 @@ def __init__(self,
         self.inputs_as_list = inputs_as_list
         self.final_weights_quantization_cfg = None
         self.final_activation_quantization_cfg = None
-        self.candidates_quantization_cfg = None
+        self.tpc_quantization_info: TPCQuantizationInfo = None
```
ofirgo: again, regarding the name of the class, I think the node's attribute name should not be connected to the tpc

model_compression_toolkit/core/common/graph/base_node.py
```diff
@@ -207,7 +186,7 @@ def _is_single_quant_mode(self, q_mode: ActivationQuantizationMode) -> bool:
             # if we have a final configuration, then we only care to check if it enables activation quantization.
             return self.final_activation_quantization_cfg.quant_mode == q_mode
 
-        q_modes = {qc.activation_quantization_cfg.quant_mode for qc in self.candidates_quantization_cfg}
+        q_modes = {qc.activation_quantization_cfg.quant_mode for qc in self.tpc_quantization_info.candidates_quantization_cfg}
```
ofirgo: why not use the property `candidates_quantization_cfg` for all the accesses to this variable?

ofirgo (commented):
Haven’t gone through everything in detail, but here are some things that jumped out to me

In addition, can you specify where do you do the patch of reloading the node with the QC information? I expected it to be in the runner but couldn't find it

model_compression_toolkit/core/common/framework_info.py
irenaby: Nothing, it's redundant. If kernel_attr exists it's a kernel op

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: We put the actual functions that perform quantization in the config, based on the method (POT etc). I think we shouldn't have implementation here, only the method enum.

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: We disable preserving quantization if it has multiple inputs or the previous op is not quantized. I kept the existing logic. I need to add description though.

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: The validation moved here. We still need the graph to contain fqc for negative shift, so set_fqc remains without validation (I think we should remove it completely, but not in this PR). BTW negative shift sets fqc for the new ops it adds so it needs access to quantization preparation functions, which is why some are public and not only the main api. Maybe we should talk about it.

irenaby:
> In addition, can you specify where do you do the patch of reloading the node with the QC information? I expected it to be in the runner but couldn't find it
In graph_prep_runner @ofirgo

model_compression_toolkit/core/common/framework_info.py
ofirgo: @elad-c please take a look, I remember talking about this and you decided to keep the method, so just to make sure it doesn't contradict what you had in mind.

model_compression_toolkit/quantization_preparation/load_fqc.py
ofirgo: I agree.
@elad-c please take a look at this as well, since you're handling the removal of the quantization methods mapping from the fw info, so it's connected

model_compression_toolkit/quantization_preparation/load_fqc.py
ofirgo: let's take this offline cause I'm still not sure what you mean

model_compression_toolkit/quantization_preparation/load_fqc.py
ofirgo: Ok let's talk about the latter offline

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: It's exactly the same logic as was previously done here
https://github.com/sony/model_optimization/blob/main/model_compression_toolkit/core/common/quantization/set_node_quantization_config.py#L198

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # qco by opset
+    # must use is_match_type for functional op in TF2.15
+    matches = [(op_type, qco) for op_type, qco in fqc.layer2qco.items() if node.is_match_type(op_type)]
+    op_types, qcos = zip(*matches) if matches else (None, None)
+    if qcos and any(qco != qcos[0] for qco in qcos[1:]):
+        raise ValueError(f'Cannot assign quantization configuration to {node} as it matches more than one op type with '
+                         f'conflicting configs: {op_types}.')
+
+    # if node matches by both filter and opset, filter takes priority (TODO irena: verify)
```
irenaby: @ofirgo Is this the intended behavior (seems reasonable)? For example can we have filter(conv with dilation) and opset(conv), or should they be mutually exclusive: filter(conv with dilation) and filter(conv without dilation)

model_compression_toolkit/core/common/framework_info.py
elad-c: There's an implementation of  it in the node, but it may still be needed when calling directly with framework layer type.

tests_pytest/conftest.py
```diff
@@ -54,6 +54,11 @@ def fw_info_mock():
     return DummyFrameworkInfo
 
 
+@fixture
+def patch_fw_info(fw_info_mock, mocker):
```
irenaby, line 58: @elad-c FYI

model_compression_toolkit/core/common/framework_info.py
irenaby: It is still redundant if called directly. You don't need a separate api to check if something is None. In any case it is not used anywhere.

irenaby: If anyone insists I don't mind adding it back to fw_info. My issue was with the redundancy in the node.

ofirgo: If it is ok by Elad it is ok by me

model_compression_toolkit/quantization_preparation/load_fqc.py
ofirgo: I see.
Are there cases that this is actually happening? where we need to disable the quantization of a preserving node because it is preserving from more than 1?
I remember there where but now when I think about it it sounds like something that we shouldn't allow
@elad-c can you weigh in?

model_compression_toolkit/quantization_preparation/load_fqc.py
ofirgo: yes this is the intended behaviour

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: I'm also not sure why we need to explicitly disable q-preserving if previous node is unquantized.

model_compression_toolkit/core/common/quantization/candidate_node_quantization_config.py
irenaby: I don't have this as a goal for the refactor, but I will have to change the way configs are read anyway so might as well keep this in mind.

model_compression_toolkit/core/common/framework_info.py
elad-c: If it's not used anywhere, delete it

model_compression_toolkit/core/common/graph/base_graph.py
```diff
@@ -578,7 +563,7 @@ def get_weights_configurable_nodes(self,
             A list of nodes that their weights can be configured (namely, has one or more weight qc candidate).
         """
         # configurability is only relevant for kernel attribute quantization
-        potential_conf_nodes = [n for n in list(self) if n.is_kernel_op]
+        potential_conf_nodes = [n for n in list(self) if n.kernel_attr]
```
yarden-yagil-sony: why not to to use the graph nodes instead of listing self? meaning as follows:
`potential_conf_nodes = [n for n in self.nodes if n.kernel_attr]`
(tested it and it works)

model_compression_toolkit/core/common/graph/functional_node.py
```diff
@@ -103,4 +103,4 @@ def is_match_type(self, _type: Type) -> bool:
 
         """
         names_match = _type.__name__ == self.type.__name__
-        return super().is_match_type(_type) or names_match
+        return names_match or super().is_match_type(_type)
```
yarden-yagil-sony, line 106: why the order of the conditions has changed?

model_compression_toolkit/core/common/mixed_precision/mixed_precision_candidates_filter.py
```diff
@@ -47,20 +40,21 @@ def filter_candidates_for_mixed_precision(graph: Graph,
         # filter out candidates activation only configurable node
         activation_configurable_nodes = [n for n in graph.get_activation_configurable_nodes()]
         for n in activation_configurable_nodes:
-            base_cfg_nbits = n.get_qco(fqc).base_config.activation_n_bits
+            base_cfg_nbits = n.quantization_cfg.base_quantization_cfg.activation_quantization_cfg.activation_n_bits
             filtered_conf = [c for c in n.candidates_quantization_cfg if
```
yarden-yagil-sony: not part of this PR, but if you can rename `filtered_conf ` to `filtered_cnfg` (or similar) that would be great :)

model_compression_toolkit/core/common/quantization/quantization_params_fn_selection.py
```diff
@@ -52,6 +52,16 @@ def get_activation_quantization_params_fn(activation_quantization_method: Quanti
     return params_fn
 
 
+weights_quant_params_fns = {
```
yarden-yagil-sony, line 55: I think "funcs" instead of "fns" is better (here and in file name)

model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/relu_bound_to_power_of_2.py
```diff
@@ -95,11 +95,11 @@ def substitute(self,
...
-            if (non_linear_node.framework_attr[HARDTANH_MIN_VAL] == 0.0) and not \
-                    (np.log2(non_linear_node.framework_attr[HARDTANH_MAX_VAL]).astype(int) -
-                     np.log2(non_linear_node.framework_attr[HARDTANH_MAX_VAL]) == 0):
-                scale_factor = non_linear_node.framework_attr[HARDTANH_MAX_VAL] / self.threshold
-                non_linear_node.functional_op.__defaults__ = (0.0, self.threshold, non_linear_node.framework_attr[INPLACE])
+            kwargs = non_linear_node.op_call_kwargs
+            if (kwargs[HARDTANH_MIN_VAL] == 0.0) and not \
+                    (np.log2(kwargs[HARDTANH_MAX_VAL]).astype(int) - np.log2(kwargs[HARDTANH_MAX_VAL]) == 0):
+                scale_factor = kwargs[HARDTANH_MAX_VAL] / self.threshold
+                non_linear_node.functional_op.__defaults__ = (0.0, self.threshold, kwargs[INPLACE])
```
yarden-yagil-sony, line 102: again not part of your changes, but if this method has already been changed I think it will be good to refactor it since it has almost identical code in each of the cases which can be simplified with additional short private methods, e.g for the hardtanh (and similarly for the relu cases) something like:
```
        def _is_hardtanh_node(non_linear_node):
            return non_linear_node.is_match_type(Hardtanh) or non_linear_node.is_match_type(hardtanh)
        ...
        elif _is_hardtanh_node(non_linear_node):
            max_val = (non_linear_node.framework_attr if non_linear_node.is_match_type(Hardtanh) else non_linear_node.op_call_kwargs)[HARDTANH_MAX_VAL] 
            is_inplace = non_linear_node.op_call_kwargs[INPLACE] if non_linear_node.is_match_type(hardtanh) else None
            if _is_zero_min_val(non_linear_node) and not (np.log2(max_val).astype(int) - np.log2(max_val) == 0):
                scale_factor = max_val / self.threshold
                set_non_linear_feilds(0.0, self.threshold, is_inplace)
```

model_compression_toolkit/core/graph_prep_runner.py
```diff
@@ -141,21 +147,26 @@ def get_finalized_graph(initial_graph: Graph,
...
-    fusing_info = FusingInfoGenerator(fqc.get_fusing_patterns()).generate_fusing_info(transformed_graph)
-    transformed_graph.fusing_info = fusing_info
-    transformed_graph.disable_fused_nodes_activation_quantization()
+    transformed_graph = load_fqc_configuration(transformed_graph, fqc)
+
+    # filter candidates per manual config
+    if bit_width_config:
+        set_manual_bitwidth_config(graph, bit_width_config)
+
+    # TODO irena: load_fqc_configuration only loads config from tpc. Previously quant_config was read as well.
```
yarden-yagil-sony, line 156: so at the end after handling these todos the `quant_config` argument of `graph_preparation_runner` will be removed?

tests/keras_tests/function_tests/test_quant_config_filtering.py
```diff
@@ -35,7 +35,9 @@
 
 get_op_set = lambda x, x_list: [op_set for op_set in x_list if op_set.name == x][0]
 
-
+# TODO irena: this tests node.filter_node_qco_by_graph which is not used anyway. What is actually used is
+#  filter_node_qco_by_graph in set_node_quantization_config which doesn't have unittests.
```
yarden-yagil-sony, line 39: so don't we want to add `set_node_quantization_config` unitests and remove this test?

tests_pytest/_test_util/graph_builder_utils.py
```diff
@@ -33,13 +32,16 @@ class DummyLayer:
 
 def build_node(name='node', canonical_weights: dict = None, final_weights: dict = None,
                qcs: List[CandidateNodeQuantizationConfig] = None,
+               sp_qc: CandidateNodeQuantizationConfig = None,
```
yarden-yagil-sony: it is not clear what `sp_qc` and `qcs` are

tests_pytest/common_tests/unit_tests/core/mixed_precision/resource_utilization_tools/test_resource_utilization_calculator.py
```diff
@@ -254,7 +254,7 @@ def test_get_a_nbits_configurable_quantization_preserving(self, graph_mock, fw_i
         {'qcs': [build_qc(42, w_attr={'foo': (4, True)}), build_qc(42, w_attr={'foo': (2, False)})]}
     ])
     def test_get_a_nbits_nonconfigurable(self, graph_mock, fw_impl_mock, node_params):
-        node = build_node(**node_params)
+        node = build_node(**node_params, disable_tpc_info_validation=True)
```
yarden-yagil-sony: why do we set `disable_tpc_info_validation=True`?
also, need to add documentation for this argument in `build_node`, and will be good to describe there when we will want to set it to True.

tests_pytest/common_tests/unit_tests/core/mixed_precision/test_mp_search_manager.py
```diff
@@ -53,8 +53,7 @@ def build_graph(fw_info_mock, w_mp: bool, a_mp: bool):
     # n3 qcs: [a4w4, a4w8, a4w2, a8w4, a8w8, a8w2]
     # n4 qcs: [w12, w4, w8]
     op_kernels = {DummyLayer1: 'foo', DummyLayer2: 'bar'}
-    fw_info_mock.get_kernel_op_attribute = lambda nt: op_kernels.get(nt, DEFAULT_KERNEL_ATTRIBUTE)
-    fw_info_mock.is_kernel_op = lambda nt: nt in op_kernels
+    fw_info_mock.get_kernel_op_attribute = lambda nt: op_kernels.get(nt)
```
yarden-yagil-sony, line 56: what is `nt` abbreviation?

model_compression_toolkit/core/common/graph/virtual_activation_weights_node.py
```diff
@@ -76,8 +75,11 @@ def __init__(self, origin_node: BaseNode, kernel_attr: str):
 
         self.name = origin_node.name + VIRTUAL_WEIGHTS_SUFFIX
 
-        self.candidates_quantization_cfg = origin_node.get_unique_weights_candidates(kernel_attr)
-        for c in self.candidates_quantization_cfg:
+        self.quantization_cfg = QuantizationConfig(
+            candidates_quantization_cfg=origin_node.get_unique_weights_candidates(kernel_attr),
+            base_quantization_cfg=None, validate=False
```
yarden-yagil-sony, line 80: can you shortly explain to me what is virtual node?
also, shouldn't `base_quantization_cfg` be initialized with one of the candidates?

model_compression_toolkit/core/common/quantization/set_node_quantization_config.py
```diff
@@ -101,6 +50,8 @@ def filter_node_qco_by_graph(node: BaseNode,
         that are compatible with next nodes supported input bit-widths.
 
     """
+    from model_compression_toolkit.quantization_preparation.load_fqc import fetch_qco_for_node
```
yarden-yagil-sony: move import to imports section

model_compression_toolkit/core/common/quantization/set_node_quantization_config.py
```diff
@@ -150,355 +102,75 @@ def filter_node_qco_by_graph(node: BaseNode,
...
+            raise ValueError(
+                f'Manually selected activation bit-width {a_nbits} is invalid for node {n}. '
+                f'Valid bit-widths: {valid_nbits}.')
+        n.quantization_cfg.candidates_quantization_cfg = candidates
+        n.quantization_cfg.base_quantization_cfg.activation_quantization_cfg.activation_n_bits = a_nbits
+
+
+# TODO irena: check coverage
+def _set_manual_weights_bitwidths(manual_weights_bitwidths: Dict[BaseNode, Dict[str, int]]):
+    def qc_attr_nbits(qc, attr, n):
```
yarden-yagil-sony, line 168: add documentation

model_compression_toolkit/core/common/quantization/set_node_quantization_config.py
```diff
@@ -150,355 +102,75 @@ def filter_node_qco_by_graph(node: BaseNode,
...
+        if attr == POS_ATTR:
```
yarden-yagil-sony: lets rename `POS_ATTR` to `POSITIONAL_ATTR`, it looks like positive_attr abbreviation :)
aslo in `pos_attributes_config_mapping` if that's not too much trouble

model_compression_toolkit/core/common/quantization/candidate_node_quantization_config.py
```diff
@@ -12,73 +12,120 @@
...
+    #  are unquantized, and handle it properly everywhere.
+    weights_quantization_cfg: Optional[NodeWeightsQuantizationConfig]
+
+
+# TODO irena: currently all code still looks at candidates_quantization_cfg as previously, so this is just an initial
+#  implementation. For now base config is completely separated from candidates (base config must be equal to one of the
+#  candidates, but we create a separate copy), and updating in place is allowed. Also we require quantization mode to
+#  be identical between all configs.
+@dataclass
+class QuantizationConfig:
```
yarden-yagil-sony: I think we need to rename it to something that make it easier to follow along the code and better differentiate it from framwork qc. We also have the same class name in `quantization_config` file..
Something like `UserQuantizationConfig` (and `user_qc`) or anything else you have in mind. For me at least it was hard to remember if qc is tpc related or not when I went through the code

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,219 @@
...
+    # TODO is not needed anymore as find min/max candidate look for a real max/min, but some tests still count on it
+    node.sort_node_candidates()
+
+    if not node.has_activation:
+        node.quantization_cfg.update_activation_quantization_mode(ActivationQuantizationMode.NO_QUANT)
+
+    _disable_unsupported_quant_preserving(node, graph)
+
+
+def fetch_qco_for_node(node: BaseNode,
```
yarden-yagil-sony: `fetch_qco_for_node` -> `fetch_fqco_for_node`

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,219 @@
...
+    Returns:
+        Updated graph.
+    """
+    graph = _set_nodes_quantization_configuration(graph, fqc)
+    graph = _set_fusion_info(graph, fqc)
+
+    return graph
+
+
+def set_quantization_configs_to_node(node: BaseNode,
```
yarden-yagil-sony, line 51: `set_quantization_configs_to_node `-> `set_framework_quantization_configs_to_node`
and similarly to the rest of the functions in this file

model_compression_toolkit/quantization_preparation/load_fqc.py
```diff
@@ -0,0 +1,219 @@
...
+    if len(prev_nodes) != 1:
+        Logger.info(f'Disabling Quantization-Preserving for node {node.name} with {len(prev_nodes)} inputs.')
+        node.quantization_cfg.update_activation_quantization_mode(ActivationQuantizationMode.NO_QUANT)
+    elif prev_nodes[0].quantization_cfg.get_activation_quant_mode() == ActivationQuantizationMode.NO_QUANT:
+        Logger.info(f'Disabling Quantization-Preserving for node {node.name} since previous node activation '
+                    f'quantization is disabled.')
+        node.quantization_cfg.update_activation_quantization_mode(ActivationQuantizationMode.NO_QUANT)
+
+
+# TODO irena copied from graph.set_fqc as is. Why does it have Keras errors?
```
yarden-yagil-sony, line 183: can we ask the relevant person and changed it if needed? so we can remove this TODO

yarden-yagil-sony (changes requested):
Looks great! 🚀 
Left some comments and questions
Also, I see some of the previously discussions here are still open, so just to make sure they are not forgotten :)

ofirgo:
@irenaby 
PR #1467, which has just been merge,d will create some conflicts with your changes. It shouldn't be complicated to fix them, but some work needs to be done for it to work correctly.
Let's talk about it offline

model_compression_toolkit/core/common/graph/functional_node.py
irenaby: No reason to always evaluate both clauses if we can do lazy evaluation.

model_compression_toolkit/core/graph_prep_runner.py
irenaby: it won't be used to "postprocess" objects created by load_fqc_configuration. If nothing else uses it in graph_prep_runner, then yes

tests/keras_tests/function_tests/test_quant_config_filtering.py
irenaby: yes, in another PR

tests_pytest/common_tests/unit_tests/core/mixed_precision/test_mp_search_manager.py
irenaby: node type

model_compression_toolkit/core/common/quantization/candidate_node_quantization_config.py
irenaby: I agree but I don't have a solution. I think UserQuantizationConfig might be a good name for the external quant config the user provides, less for a config based on TPC (even though the user provides it as well), but it's a user API change. 
I'm going to change this to NodeQuantizationConfig, at least it will provide the scope. I'm open to other suggestions.

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: In another PR

model_compression_toolkit/core/common/graph/virtual_activation_weights_node.py
irenaby: Virtual node is used to define mixed precision candidates when BOPS are restricted. BOPS count depends both on kernel bitwidth and input activation bitwidth, but MP candidates contain kernel and output activation, and we always evaluate a single candidate. So we break a linear layer into weights virtual node and activation virtual node, and then recombine the activation node with subsequent (instead of previous) weight node. Call for more...
We only build virtual nodes to define candidates for MP, so base config of a virtual node should never be used, so we allow None in this specific case.

model_compression_toolkit/core/pytorch/graph_substitutions/substitutions/relu_bound_to_power_of_2.py
irenaby: I agree it's horrible, but won't address it now

tests_pytest/_test_util/graph_builder_utils.py
irenaby: changed sp_qc to base_qc. 
qcs is used ubiquitously for quantization candidates

tests_pytest/common_tests/unit_tests/core/mixed_precision/resource_utilization_tools/test_resource_utilization_calculator.py
irenaby: By default NodeQuantizationConfig validates that all candidates have the same quantization mode per attr. The logic that is tested doesn't make this assumption so it's tested with some candidates quantized and some unquantized, so need to disable the validation. 
Added comment here and description in build_node (and changed the name of the flag as TPCInfo was renamed)

model_compression_toolkit/core/common/quantization/set_node_quantization_config.py
irenaby: It creates circular import. This function is going to be moved in one of the next PRs which will resolve it

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: fqc is framework quantization capabilities, qco is quantization config options. The names are super confusing, especially if its the first time you see them, but I think fqco is even more confusing. Renaming to fetch_qc_options_for_node.

model_compression_toolkit/quantization_preparation/load_fqc.py
irenaby: i think the bad name is actually FQC, it implies that quantization capabilities are per framework, which is misleading

model_compression_toolkit/core/common/quantization/quantization_params_fn_selection.py
irenaby: keeping as is, fn is pretty standard for function

PR #1480, opened 2025-06-30
Remove QuantizationConfig fields patch from NodeActivationQuantizationConfig

irenaby:
## Pull Request Description:
Remove the intermediate patch that injected fields from QuantizationConfig back to NodeActivationQuantizationConfig.
Use QuantizationConfig directly in all places that read it from the node candidates.
Remove min_threshold from TrainableQuantizerActivationConfig.


tests/keras_tests/feature_networks_tests/feature_networks/network_editor/edit_error_method_test.py
```diff
@@ -1,70 +0,0 @@
...
-    ChangeCandidatesActivationQuantConfigAttr
-from model_compression_toolkit.core.common.network_editors.node_filters import NodeTypeFilter
-from tests.keras_tests.feature_networks_tests.base_keras_feature_test import BaseKerasFeatureNetworkTest
-from tests.keras_tests.utils import get_layers_from_model_by_type
-
-keras = tf.keras
-layers = keras.layers
-
-
-class EditActivationErrorMethod(BaseKerasFeatureNetworkTest):
```
yarden-yagil-sony, line 36: Why this test is deleted? and if it's due to changes, don't we want to create a new version of it with adaptations?

model_compression_toolkit/trainable_infrastructure/common/trainable_quantizer_config.py
```diff
@@ -44,7 +44,6 @@ def __init__(self,
                  activation_n_bits: int,
                  activation_quantization_params: Dict,
                  enable_activation_quantization: bool,
-                 min_threshold: float,
```
yarden-yagil-sony, line 47: why is it removed from the quantizer config?

irenaby: There is also a MIN_THRESHOLD const and in some places the const was used directly and in some places it came from the config (which has the same default value and most probably no one ever used a non-default value). So it was decided to just consistently use the const and remove it from the config.

tests/keras_tests/feature_networks_tests/feature_networks/network_editor/edit_error_method_test.py
irenaby: The whole edit rule was removed as it edited a field that is no more part of the node activation config, so there is nothing to replace with

PR #1486, opened 2025-07-01
Node weights and node weights attr quantization configs clean-up - remove temporary patch for fields from QuantizationConfig

irenaby:
## Pull Request Description:
*Breaking change:* removed min_threshold from QuantizationConfig.
Remove the temporary patch that injected fields previously read from QuantizationConfig back into NodeWeightsQuantizationConfig and WeightsAttrQuantizationConfig.
Pass QuantizationConfig or its relevant field(s) instead of getting it from the node.
Second moment correction: for now keep weights_second_moment_correction flag in NodeWeightsQuantizationConfig, as it is used to mark the nodes for which to apply the correction and remove the reconstructed BN, but instead of reading from QuantizationConfig, set it in BN reconstruction substitution.
Bias correction:  remove weights_bias_correction flag from NodeWeightsQuantizationConfig. bias_corrected field is still injected into the config, use it instead to determine whether bias correction should be applied. 

### DIFF
--- a/model_compression_toolkit/core/common/quantization/node_quantization_config.py
+++ b/model_compression_toolkit/core/common/quantization/node_quantization_config.py
@@ -18,7 +18,6 @@ from enum import Enum, auto
 from model_compression_toolkit.core.common.framework_info import ChannelAxisMapping
 from model_compression_toolkit.logger import Logger
 
-from model_compression_toolkit.core.common.quantization.quantization_config import QuantizationConfig
 from model_compression_toolkit.target_platform_capabilities.constants import POSITIONAL_ATTR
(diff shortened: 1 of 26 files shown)


model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -63,6 +63,7 @@ def set_quant_config_attr(self, config_parameter_name: str, config_parameter_val
         if hasattr(self, config_parameter_name):
             setattr(self, config_parameter_name, config_parameter_value)
         else:
+            raise AttributeError(config_parameter_name)
```
elad-c: Why not delete the logger warning after this error?

elad-c (approved):

elad-c (approved):

PR #1487, opened 2025-07-02
Apply activation quantization parameters selection.

gouda-youichi:
Apply activation quantization parameters selection (Stage3-4)

## Pull Request Description:
Apply activation quantization parameters selection (Stage3-4)


model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+from __future__ import annotations
```
ofirgo: why do you need this import?

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -140,15 +143,17 @@ def fln_quantization(self):
...
         """
          Set a quantization parameter for the node's activation.
 
         Args:
             activation_params: Dictionary that contains weight quantization params.
+            node: node in a graph that represents the model.
 
         """
-        assert self.quant_mode == ActivationQuantizationMode.QUANT
+        assert node.is_activation_quantization_enabled() or node.is_fln_quantization()
```
ofirgo: no need to add node to the function's arguments, the node's check `is_fln_quantization` is actually inspecting the node's candidates to decide if this an FLN. This function `set_activation_quantization_param` is already a candidate class function, so you can just check the class `quant_mode`

tests_pytest/pytorch_tests/unit_tests/core/common/quantization/quantization_params_generation/test_calculate_quantization_params.py
```diff
@@ -0,0 +1,232 @@
...
+from model_compression_toolkit.core.common.framework_info import set_fw_info, get_fw_info
+
+from model_compression_toolkit.core.pytorch.pytorch_implementation import PytorchImplementation
+from model_compression_toolkit.core.common.collectors.statistics_collector import StatsCollector
+from model_compression_toolkit.target_platform_capabilities.constants import KERNEL_ATTR, WEIGHTS_N_BITS
+from mct_quantizers import QuantizationMethod
+
+from model_compression_toolkit.core.common.framework_info import ChannelAxisMapping
+
+class TestCalculateQuantizationParams:
```
ofirgo: As we discussed in the previous PR, try to simplify the unit tests by doing the following:
1. Separate the one big test into several tests, each for testing specific functionality or use case. You can use pytest parametrization or create several tests under the same class.
2. Use mocks for utility functionality that is not part of the test.
3. Add some description for the functions to explain your intentions and the expected behaviour

ofirgo (changes requested):

model_compression_toolkit/core/common/quantization/node_quantization_config.py
gouda-youichi: If this import is missing, the following import will result in a "not defined" error:

```
if TYPE_CHECKING:
    from model_compression_toolkit.core.common import BaseNode
```

Also, if TYPE_CHECKING: is missing, a circular reference error will occur.

ofirgo: You can remove the BaseNode import since you don't need to pass a node to the function.
After removing the BaseNode import, try to remove the annotations import, you should be able to do so.

gouda-youichi: I removed the BaseNode import.

tests_pytest/pytorch_tests/unit_tests/core/common/quantization/quantization_params_generation/test_calculate_quantization_params.py
gouda-youichi: I simplified the test.

tests_pytest/common_tests/unit_tests/core/quantization/quantization_params_generation/test_calculate_quantization_params.py
```diff
@@ -0,0 +1,118 @@
...
+        graph.node_to_out_stats_collector = dict()
+        for n in graph.nodes():
+            n.prior_info = NodePriorInfo()
+
+            graph.node_to_out_stats_collector[n] = StatsCollector(init_min_value=0.0, init_max_value=1.0, out_channel_axis=0)
+            graph.node_to_out_stats_collector[n].hc._n_bins = 3
+            graph.node_to_out_stats_collector[n].hc._bins = np.array(data)
+            graph.node_to_out_stats_collector[n].hc._counts = np.array([1, 1])
+
+        quant_config = QuantizationConfig()
```
ofirgo: if this is the same (default QuantizationConfig) for all test cases, it is clearer to just initialized it in the test and not as part of the graph setup

ofirgo (approved):

PR #1490, opened 2025-07-02
Graph prepare refactor

reuvenperetz:
## Pull Request Description:
Extract the parts of graph building from the framework model to an outer package called "graph_builder".
BaseGraphBuilder defines API methods for converting the model to a graph and transforming it using the fundamental substitutions. 


model_compression_toolkit/graph_builder/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2025 Sony Semiconductor Israel, Inc. All rights reserved.
```
ofirgo, line 1: I think that graph_preparation is a better name for the package

model_compression_toolkit/graph_builder/common/base_graph_builder.py
```diff
@@ -0,0 +1,112 @@
...
+            relu_bound_to_power_of_2 (bool): If True, replaces ReLU activation bounds with nearest power-of-2 (for HW-friendliness).
+
+        Returns:
+            Graph: The transformed, simplified computational graph.
+        """
+        graph = self.convert_model_to_graph(model, representative_dataset)
+        if tensorboard_writer is not None:
+            tensorboard_writer.add_graph(graph, 'initial_graph')
+
+        if fqc:
```
ofirgo, line 67: I don't think that fqc should be an argument to this module, since it is quantization-related.
If we must still store it in the graph, then it supposed to be the "quantization_preparation" responsibility

model_compression_toolkit/graph_builder/common/base_graph_builder.py
```diff
@@ -0,0 +1,112 @@
...
+    Subclasses must implement methods to:
+    1. Convert a model to a graph representation.
+    2. Apply optional graph simplifications and preprocessing.
+    """
+
+    def build_graph(self,
+                    model: Any,
+                    representative_dataset: Callable = None,
+                    fqc: FrameworkQuantizationCapabilities = None,
+                    tensorboard_writer: TensorboardWriter = None,
```
ofirgo, line 41: Several things regarding the arguments:

1. Move tensor board  to be the last argument.
2. Consider maybe gathering the 3 "structure-altering" flags in a dedicated config?
3. Do we need the default values to not "break" API/tests? if not, then remove the default.
4. Change the typehint of arguments that accept `None` to `Optional[]`

model_compression_toolkit/graph_builder/pytorch/convert_pytorch_model_to_graph.py
```diff
@@ -0,0 +1,45 @@
...
+
+import copy
+from typing import Callable
+from model_compression_toolkit.core.pytorch.utils import torch_tensor_to_numpy, to_torch_tensor
+from model_compression_toolkit.core.common import Graph
+from model_compression_toolkit.graph_builder.pytorch.reader.reader import model_reader
+import torch
+
+
+def convert_pytorch_model_to_graph(model: torch.nn.Module,
```
ofirgo, line 24: Why split it to a separate file and not include the implementation as part of `PytorchGraphBuilder`'s `convert_model_to_graph` method? (same for keras)

model_compression_toolkit/graph_builder/pytorch/transform_pytorch_graph.py
```diff
@@ -0,0 +1,111 @@
...
+    FunctionalConvSubstitution
+from model_compression_toolkit.core.pytorch.graph_substitutions.substitutions.relu_bound_to_power_of_2 import \
+    ReLUBoundToPowerOfTwo
+from model_compression_toolkit.core.pytorch.graph_substitutions.substitutions.remove_identity import RemoveIdentity
+from model_compression_toolkit.core.pytorch.graph_substitutions.substitutions.reshape_with_static_shapes import \
+    ReshapeWithStaticShapes
+from model_compression_toolkit.core.common.substitutions.apply_substitutions import substitute
+
+
+def transform_pytorch_graph(graph: Graph,
```
ofirgo, line 55: same question, why not as part of the pytorch builder class?

model_compression_toolkit/ptq/keras/quantization_facade.py
```diff
@@ -154,7 +154,8 @@ def keras_post_training_quantization(in_model: Model,
                                                                 fw_impl=fw_impl,
                                                                 fqc=framework_platform_capabilities,
                                                                 target_resource_utilization=target_resource_utilization,
-                                                                tb_w=tb_w)
+                                                                tb_w=tb_w,
+                                                                fw_graph_builder=KerasGraphBuilder())
```
ofirgo, line 158: Please add a todo either here (in each facade) or in the core_runner that this should eventually be initialized and run outside the runner.

tests/keras_tests/graph_tests/test_graph_reading.py
```diff
@@ -21,9 +21,9 @@
 from tensorflow.keras.applications.nasnet import NASNetMobile
 from tensorflow.keras.applications.resnet import ResNet50
 from tensorflow.keras.applications.xception import Xception
-from model_compression_toolkit.core.keras.reader.reader import model_reader
 from model_compression_toolkit.core.common.framework_info import set_fw_info
 from model_compression_toolkit.core.keras.default_framework_info import KerasInfo
+from model_compression_toolkit.graph_builder.keras.convert_keras_model_to_graph import convert_keras_model_to_graph
 
 
 class TestGraphReading(unittest.TestCase):
```
ofirgo, line 29: As part of the PR where you impement tests for the new module (GraphBuilder), also migrate/remove this test and any other test that is meant to verify the graph building and transformation

model_compression_toolkit/graph_builder/common/base_graph_builder.py
```diff
@@ -0,0 +1,112 @@
...
+                    linear_collapsing: bool = True,
```
ofirgo, line 42: what about the substitutions in `get_substitutions_post_statistics_collection`? aren't they graph preparation related? if so, need to think if they need to be part of this module or somewhere else

model_compression_toolkit/graph_builder/common/base_graph_builder.py
reuvenperetz: I started without it, but the relu_to_pot substitution requires it, which is why I added it eventually. 
I'll do a different PR to remove this requirement first.

model_compression_toolkit/graph_builder/common/base_graph_builder.py
reuvenperetz: About 2: That's how I started, but I eventually thought it would cause unnecessary maintenance. I can write it back.

model_compression_toolkit/graph_builder/common/base_graph_builder.py
reuvenperetz: They depend on the quantization params (this is why they are used today only after loading the fqc, and computing quantization params). So I do not think they should be part of it...

model_compression_toolkit/ptq/keras/quantization_facade.py
reuvenperetz: I was trying to avoid duplicating code. Do you see any benefit in moving it outside rather than passing the framework class builder?

model_compression_toolkit/graph_builder/common/base_graph_builder.py
ofirgo: So depends on what the unnecessary maintanance is.
Normally, I wouldn't suggest to just have a config as an argument that is immidiatly flattend in the class init, but in this case they are all substitution bollean flags so it feels right to tie them together in a "which substitutions to run" config variable

model_compression_toolkit/ptq/keras/quantization_facade.py
ofirgo: Eventually, the runner is going to die. The point of separating each logical block from the core into a new module is exactly this - to not have 1 backbone that you have to run in order to get through anything in the MCT.
So, the endgame is to have each facade call the 4-5 modules that it needs to - construct graph, put quantization info, run some optimization, run any enhancement that is part of the facade's algorithmic framework.

ofirgo: To summarize my long answer - yes, I see benefites in moving it outside of the runner

model_compression_toolkit/graph_builder/common/base_graph_builder.py
ofirgo: Let's talk about this offline, I want us to take another look at this an figure out where they should land.

From a quick check this is what I think can work:
1. softmax shift - not quantization related, just need to happen after statistic collection, so maybe put it in this module but not part of the "graph_builder" flow (as a separate api that can be called)
2. concat threshold - this needs to happen after parameters selection, so it is like any other "quantization enhancement" substitution that we have (bc, snc...)
3. input scaling - first, it is not available in pytorch, maybe we can earase it when limiting keras support. even if not - it is like the concat threshold - a quantization enhancement

PR #1495, opened 2025-07-09
Invalidate irrelevant configuration in node quantization configs

irenaby:
## Pull Request Description:
Un-set configuration in node quantization configs if it is or becomes irrelevant. This was we don't hold invalid params, prevent using them by mistake, and removing redundant candidates becomes much easier (given that we don't want to completely restructure the configs at this point). NodeQuantizationConfig removes duplicate candidates automatically.
Replace old tests for nodes configs.


tests_pytest/common_tests/unit_tests/core/quantization/test_node_weights_quantization_config.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def assert_unset_attr_config(cfg: WeightsAttrQuantizationConfig):
+    assert cfg.weights_quantization_method is None
+    assert cfg.weights_n_bits == 0
+    assert cfg.weights_per_channel_threshold is None
+    assert cfg.weights_channels_axis is None
+
+
+class TestWeightsQuantizationConfig:
```
irenaby, line 88: No change from here to line 224 (file was moved)

model_compression_toolkit/core/common/quantization/candidate_node_quantization_config.py
```diff
@@ -46,6 +46,20 @@ class NodeQuantizationConfig:
 
     validate: InitVar[bool] = True
 
+    def __post_init__(self, validate=True):
```
elad-c, line 49: Can `__post_init__` take input arguments?

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -122,9 +148,29 @@ def set_activation_quantization_param(self,
...
+            self.set_quant_mode(value)
+        else:
+            if self.quant_mode in self._no_cfg_modes:
+                raise ValueError(f'Cannot set attribute {attr_name} for activation with disabled quantization.')
+            super().set_quant_config_attr(attr_name, value)
+
+    def _unset(self):
+        """ Unset activation quantization fields to None. """
+        self.activation_quantization_method = None
+        self.activation_n_bits = 0
```
elad-c, line 172: Why not `None`?

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -320,19 +392,22 @@ def get_attr_config(self, attr_name: 'WeightAttrT') -> WeightsAttrQuantizationCo
 
         return attr_cfg
 
-    def set_attr_config(self, attr_name: 'WeightAttrT', attr_qc: WeightsAttrQuantizationConfig):
+    def set_attr_config(self, attr_name: 'WeightAttrT', attr_qc: WeightsAttrQuantizationConfig, force=False):
```
elad-c, line 395: typehint

model_compression_toolkit/core/common/quantization/node_quantization_config.py
```diff
@@ -405,25 +477,35 @@ def set_quant_config_attr(self, config_parameter_name: str, config_parameter_val
...
-                Logger.critical(f"Weights attribute {attr_name} could not be found to set parameter {config_parameter_name}.")
+                                                                             config_parameter_value)
+            return
+
+        if not self.has_attribute_config(attr_name):
+            raise ValueError(
+                f"Weights attribute {attr_name} could not be found to set parameter {config_parameter_name}.")
+
+        attr_cfg = self.get_attr_config(attr_name)
+        if config_parameter_name == 'enable_weights_quantization':
```
elad-c, line 491: Can replace str with constant?

model_compression_toolkit/core/common/statistics_correction/apply_activation_bias_correction_to_graph.py
```diff
@@ -71,7 +71,8 @@ def _apply_activation_bias_correction_to_node(node: BaseNode,
         node.final_weights_quantization_cfg.set_attr_config(fw_impl.constants.BIAS,
                                                             WeightsAttrQuantizationConfig(
                                                                 AttributeQuantizationConfig(
-                                                                    enable_weights_quantization=False)))
+                                                                    enable_weights_quantization=False)),
+                                                            force=True)
```
elad-c, line 75: Why is `force` needed? Add comment for it

model_compression_toolkit/core/common/statistics_correction/apply_bias_correction_to_graph.py
```diff
@@ -68,4 +68,5 @@ def _apply_bias_correction_to_node(node: BaseNode,
         node.framework_attr[fw_impl.constants.USE_BIAS] = True  # Mark the use_bias attribute of the node.
         node.final_weights_quantization_cfg.set_attr_config(fw_impl.constants.BIAS,
                                                             WeightsAttrQuantizationConfig(AttributeQuantizationConfig(
-                                                                enable_weights_quantization=False)))
+                                                                enable_weights_quantization=False)),
+                                                            force=True)
```
elad-c, line 72: Same about force

model_compression_toolkit/core/common/statistics_correction/apply_activation_bias_correction_to_graph.py
irenaby: Because the attribute doesn't exist in the config. Without force it will raise an error. Will add a comment.

model_compression_toolkit/core/common/quantization/node_quantization_config.py
irenaby: It's not a const, it's a name of an attribute in the same class, it's meaningless to define it as const

model_compression_toolkit/core/common/quantization/node_quantization_config.py
irenaby: There several places where we look for max/min candidate or sort based on nbits whether quantization is enabled or not, which still work with 0 but not with None. I wanted to minimize changes elsewhere and 0 is clearly not a real value. That is the reason, but it's probably not a good enough reason.

PR #1528, opened 2025-10-14
Fixed pydantic schema to enable support Pydantic 2.12

alysenko-nn:
## Pull Request Description:
MCT doesn't work correctly with `Pydantic 2.12`. Root cause is the old-style validator format, which was deprecated in v2.12. Validators were updated to v2 style and work without any issues with the latest (2.12) version


kmk142789 (approved):

YingYang-YY:
Hi, thank you for raising the issue. We added the version restriction "pydantic>=2.0,<2.12.0" from MCT 2.4.4.

Issue #1451, opened 2025-05-26
IndexError: tuple index out of range


eladc-git:
### Issue Type

Bug

### Source

source

### MCT Version

2.3.0

### Python version

3.11

### Describe the issue

MCT quantization crashes on ModelCollector when the tensor is scalar (1 element)

### Expected behaviour

MCT quantization will pass

### Code to reproduce the issue

```python
import numpy as np
import model_compression_toolkit as mct
import torch
import torch.nn as nn
import edgemdt_tpc


class Model(nn.Module):
    def __init__(self):
        super(Model, self).__init__()
        self.conv = nn.Conv2d(3, 3, kernel_size=3, padding=1)
        self.bias = nn.Parameter(torch.tensor([-1.0]))
        self.scale = nn.Parameter(-1.0 * torch.ones([]))

    def forward(self, x):
        x = self.conv(x)
        y = torch.exp(self.scale) * x + self.bias
        return y


float_model = Model()
#a = float_model(torch.randn((1, 3, 224, 224)))

def representative_data_gen():
    yield [np.random.random((1, 3, 224, 224))]


tpc = edgemdt_tpc.get_target_platform_capabilities("4.0")
quantized_model, _ = mct.ptq.pytorch_post_training_quantization(float_model,
                                                                representative_data_gen=representative_data_gen,
                                                                target_platform_capabilities=tpc)
```

### Log output

2025-05-26 09:53:54.049632: I tensorflow/core/util/port.cc:113] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
WARNING:tensorflow:From C:\Users\1000322838\PycharmProjects\model_optimization\venv\lib\site-packages\keras\src\losses.py:2976: The name tf.losses.sparse_softmax_cross_entropy is deprecated. Please use tf.compat.v1.losses.sparse_softmax_cross_entropy instead.

C:\Users\1000322838\PycharmProjects\model_optimization\venv\lib\site-packages\torchvision\io\image.py:13: UserWarning: Failed to load image Python extension: '[WinError 127] The specified procedure could not be found'If you don't plan on using image functionality from `torchvision.io`, you can ignore this warning. Otherwise, there might be something wrong with your environment. Did you have `libjpeg` or `libpng` installed before building `torchvision` from source?
  warn(
[...]

PR #1660, opened 2026-02-16
Support for scalars in statistics collector

kawakami-masaki0:
## Pull Request Description:
This PR addresses the following https://github.com/SonySemiconductorSolutions/mct-model-optimization/issues/1451
In PyTorch, if the number of dimensions of the activation tensor is less than 2, an error occured in the statistics collection process.

[Cause]
1. When converting to a Numpy array, the scalar remains 0-dimensional, so the tensor's dimension `x.shape` cannot be obtained.
2. The default value of `out_channel_axis` is 1, which means that axes cannot be set correctly for tensors with dimensions less than 2.

[Solution]
1. Support for scalars in `torch_tensor_to_numpy`.
2. Modified `out_channel_axis` to -1 if the number of dimensions of the activation tensor < 2.

### DIFF
--- a/tests_pytest/pytorch_tests/e2e_tests/test_quantization_for_1d_tensor.py
+++ b/tests_pytest/pytorch_tests/e2e_tests/test_quantization_for_1d_tensor.py
@@ -27,68 +27,70 @@ class Model(nn.Module):
         self.name = name
         self.conv = nn.Conv2d(3, 3, kernel_size=3, padding=1)
         self.relu = nn.ReLU()
+        self.tensor = nn.Parameter(2.0 * torch.ones([1])) # 1D tensor
 
     def forward(self, x):
         x = self.conv(x)
         x = self.relu(x)
-        x = torch.reshape(x, (-1,))
 
         if self.name == 'add':
-            y = torch.add(x, 1)
+            const = torch.add(self.tensor, 1)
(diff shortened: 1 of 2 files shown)


tests_pytest/pytorch_tests/unit_tests/core/test_create_stats_collector_for_node.py
```diff
@@ -0,0 +1,93 @@
...
+    pass
+
+class DummyLayer:
+    pass
+
+@pytest.fixture
+def fw_info_mock():
+    fw_info = Mock()
+    fw_info.out_channel_axis_mapping = DefaultDict({Conv2D: 1, Linear: -1, ConvTranspose2d: 1}, 1)
+    return fw_info
```
kkawa14, line 37: It looks like out_channel_axis_mapping is only being tested with PyTorch, 
but from a coverage perspective we should also verify the Keras pattern.
If there are no issues there, we don’t need integration or end-to-end tests.

kawakami-masaki0: Added test for Keras.
We've confirmed that the out_channel_axis remains -1, so we don't need to create integration or end-to-end tests.

